PR #5, opened 2022-10-12
New rule that detects space before parentheses

realmarv:


commitlint.config.ts
```diff
@@ -207,6 +207,28 @@ module.exports = {
...
+                        if (parenIndex >= 1){
+                            if (headerStr[parenIndex - 1] === ' ') {
+                                offence = true;
+                            }
+                        }    
+                    }
+
+                    return [
+                        !offence,
+                        `Please watch out for leading or ending trailing whitespace`
```
knocte, line 229: @realmarv well done so far! but we made a mistake here ^ (copy+paste problem hehehe)

commitlintplugins.test.ts
```diff
@@ -205,3 +205,23 @@ test('type-space-after-comma2', () => {
...
+    let typeSpaceBeforeParen2 = runCommitLintOnMsg(commitMsgWithNoSpaceBeforeParen);
+    expect(typeSpaceBeforeParen2.status).toBe(0);
+});
+
+
+test('type-space-before-paren3', () => {
+    let commitMsgWithNoSpaceBeforeParen = "(bar): bla bla bla";
+    let typeSpaceBeforeParen3 = runCommitLintOnMsg(commitMsgWithNoSpaceBeforeParen);
+    expect(typeSpaceBeforeParen3.status).toBe(0);
+});
```
knocte, line 227: @realmarv nit: also please add EOL before EOF here ^

knocte (changes requested):
See, I added 2 comments to address. After you fix those, I'll merge the PR.

commitlint.config.ts
```diff
@@ -226,7 +226,7 @@ module.exports = {
 
                     return [
                         !offence,
-                        `Please watch out for leading or ending trailing whitespace`
+                        `Please watch out for whitespaces before parentheses`
```
knocte, line 229: @realmarv mmm, this message might be confusing to the user; think about this case! he writes the commit message `Frontend (SendPage): fix crash (thread issue)`, then she will remove all the spaces before parentheses; but we only care about the first one (because it is inside the scope).

commitlint.config.ts
```diff
@@ -207,6 +207,28 @@ module.exports = {
...
+                        if (parenIndex >= 1){
+                            if (headerStr[parenIndex - 1] === ' ') {
+                                offence = true;
+                            }
+                        }    
+                    }
+
+                    return [
+                        !offence,
+                        `Please watch out for whitespaces before parentheses`
```
knocte, line 229:
```suggestion
                        `No need to use space before parentheses in the area/scope/sub-area section of the title`
```

knocte:
@realmarv Unfortunately, because I merged the other PR, now this PR has conflicts and cannot be merged, so it needs to be rebased. Please don't use the "git merge" command to do this, you should use "git rebase". If you need help with this, ask in channel please :)

commitlint.config.ts
realmarv: Done!

commitlintplugins.test.ts

commitlint.config.ts
realmarv: Done!

commitlint.config.ts
realmarv: Done!

realmarv:
> @realmarv Unfortunately, because I merged the other PR, now this PR has conflicts and cannot be merged, so it needs to be rebased. Please don't use the "git merge" command to do this, you should use "git rebase". If you need help with this, ask in channel please :)

Done!

commitlint.config.ts
```diff
@@ -28,17 +28,17 @@ module.exports = {
...
         'footer-max-line-length': [RuleStatus.Error, 'always', 150],
         'header-max-length': [RuleStatus.Error, 'always', 50],
+        'prefer-slash-over-backslash': [RuleStatus.Error, 'always'],
         'subject-full-stop': [RuleStatus.Error, 'never', '.'],
         'type-empty': [RuleStatus.Warning, 'never'],
         'type-space-after-colon': [RuleStatus.Error, 'always'],
         'subject-lowercase': [RuleStatus.Error, 'always'],
         'type-space-after-comma': [RuleStatus.Error, 'always'],
         'trailing-whitespace': [RuleStatus.Error, 'always'],
-        'prefer-slash-over-backslash': [RuleStatus.Error, 'always'],
```
knocte, line 37: @realmarv when rebasing, you deleted this by mistake; this line needs to be reinstated before I can merge this PR

realmarv: It's in line 31. I thought rules should be ordered alphabetically, but now that I know it's based on their priority, should I put it at the end of the list?

knocte: we were writing tests in alphabetical level; then in another PR I pointed out that rules could be in alphabetical level too; but we never talked about configuration of rules being in alphabetical order

knocte: typo: level -> order

realmarv: Done!

commitlintplugins.test.ts
```diff
@@ -219,3 +219,23 @@ test('type-space-after-comma2', () => {
...
+
+test('type-space-before-paren2', () => {
+    let commitMsgWithNoSpaceBeforeParen = "foo(bar): bla bla bla";
+    let typeSpaceBeforeParen2 = runCommitLintOnMsg(commitMsgWithNoSpaceBeforeParen);
+    expect(typeSpaceBeforeParen2.status).toBe(0);
+});
+
+
+test('type-space-before-paren3', () => {
+    let commitMsgWithNoSpaceBeforeParen = "(bar): bla bla bla";
```
knocte, line 238: last thing to do in this PR: as we found this case above as an edge case for this rule, because it's an invalid title (as it has sub-area but no area), but shouldn't be covered by this rule, let's add a TODO item in the list that explains we need a rule for this

realmarv: Should I remove this test? Because as you said, it's not a correct title, but in this test, we expect the status code to be 0.

knocte: good question, maybe we should remove it (or rather rename it) when we implement the new rule that will take care of titles with sub-area but without area

realmarv: Right

PR #7, opened 2022-10-12
New rule that forces using slash over backslash in area/scope

realmarv:


knocte:
If this PR supersedes another one, please close the old one (when closing, add a comment linking to the new PR).

commitlint.config.ts
```diff
@@ -207,6 +207,25 @@ module.exports = {
                         `Please watch out for leading or ending trailing whitespace`
                     ];
                 },
+
+                'prefer-slash-over-backslash': ({header}: {header:any}) => {
```
knocte, line 211: @realmarv please move this plugin to not be the last one (try to make it alphabetical, even though they are not really following proper order now, but at least let's not make it worse :) )

commitlintplugins.test.ts
```diff
@@ -73,6 +73,33 @@ test('body-max-line-length6', () => {
...
+    let preferSlashOverBackslash3 = runCommitLintOnMsg(commitMsgWithSlash);
+    expect(preferSlashOverBackslash3.status).toBe(0);
+});
+
+
+test('prefer-slash-over-backslash4', () => {
+    let commitMsgWithBackslash = "\\bar: bla bla bla";
+    let preferSlashOverBackslash4 = runCommitLintOnMsg(commitMsgWithBackslash);
+    expect(preferSlashOverBackslash4.status).not.toBe(0);
+});
```
knocte, line 100: @realmarv not sure I see the value of the last 2 tests, let's just remove them because they add more confusion than anything

knocte:
Good work, merging.

PR #8, opened 2022-10-12
New plugin to prevent adding a dot at the end of the title

realmarv:


knocte:
I realised there's already a rule for this: `subject-full-stop`. We need to investigate why it's not working?

realmarv:
> I realised there's already a rule for this: `subject-full-stop`. We need to investigate why it's not working?

I think the problem with `subject-full-stop` is that the conventional-changelog-conventionalcommits that commitlint is using for parsing its commits, assumes the format of the commit header is `type(scope): subject`, and for some reason it can't break this header: `foo/bar: bla bla blah.` and the subject field will become null and `subject-full-stop` will not work. These two regex patterns are being used in conventionalcommits:
```
headerPattern: /^(\w*)(?:\((.*)\))?!?: (.*)$/
breakingHeaderPattern: /^(\w*)(?:\((.*)\))?!: (.*)$/,
```
I also don't understand the reason for "!" before the colon in the second pattern.

realmarv:
> > I realised there's already a rule for this: `subject-full-stop`. We need to investigate why it's not working?
> 
> I think the problem with `subject-full-stop` is that the conventional-changelog-conventionalcommits that commitlint is using for parsing its commits, assumes the format of the commit header is `type(scope): subject`, and it can't break this header: `foo/bar: bla bla blah.` bacause the regex pattern it has used for type is `(\w*)` which doesn't include "/" character. As a result the subject field will become null and `subject-full-stop` will not work. These two regex patterns are being used in conventionalcommits:
> 
> ```
> headerPattern: /^(\w*)(?:\((.*)\))?!?: (.*)$/
> breakingHeaderPattern: /^(\w*)(?:\((.*)\))?!: (.*)$/,
> ```
> 
> I also don't understand the reason for "!" before the colon in the second pattern.

I added two tests here: https://github.com/realmarv/conventions/commits/notAllowingEndingDotInTitle-squashed
As you can see the first one that is without "/" character before colon has passed but the second test hasn't passed.

realmarv:
> > I realised there's already a rule for this: `subject-full-stop`. We need to investigate why it's not working?
> 
> I think the problem with `subject-full-stop` is that the conventional-changelog-conventionalcommits that commitlint is using for parsing its commits, assumes the format of the commit header is `type(scope): subject`, and it can't break this header: `foo/bar: bla bla blah.` bacause the regex pattern it has used for type is `(\w*)` which doesn't include "/" character. As a result the subject field will become null and `subject-full-stop` will not work. These two regex patterns are being used in conventionalcommits:
> 
> ```
> headerPattern: /^(\w*)(?:\((.*)\))?!?: (.*)$/
> breakingHeaderPattern: /^(\w*)(?:\((.*)\))?!: (.*)$/,
> ```
> 
> I also don't understand the reason for "!" before the colon in the second pattern.

To fix the problem I think one way is to pass the following regex patterns instead of the above patterns to the conventionalcommits as parser options from commitlint, to include the "/" character in the area as well:
```
headerPattern: /^([\w\/]*)(?:\((.*)\))?!?: (.*)$/,
breakingHeaderPattern: /^([\w\/]*)(?:\((.*)\))?!: (.*)$/,
```

knocte:
Another testcase that should fail and doesn't: `foo.bar: bla blah.`

knocte:
Superseded by https://github.com/nblockchain/conventions/pull/70

PR #9, opened 2022-10-13
Add an exception to subject-lowercase rule to allow PascalCase after colon in title

realmarv:


commitlintplugins.test.ts
```diff
@@ -5,6 +5,13 @@ function runCommitLintOnMsg(inputMsg: string) {
 }
 
 
+test('allow-PascalCase-after-colon-in-title', () => {
```
knocte, line 8: @realmarv this rule is about the rule 'subject-lowercase', so let's rename this test and move it next to where the other subject-lowercase tests are

commitlintplugins.test.ts
```diff
@@ -5,6 +5,13 @@ function runCommitLintOnMsg(inputMsg: string) {
...
+    let commitMsgWithPascalCaseAfterColon = "foo/bar: PascalCase bla blah";
```
knocte, line 9: 2 nits here:
- The simpler the test, the better. In this sense, there's no need to add a sub-area, so let's replace `foo/bar` with `foo` (this way, we also reduce the possible overlap between tests; i.e. imagine that some future dev inverts, by mistake, the rule that prefers slash over backslash, then the consequence of it would be to have this test broken even if it's unrelated to that; so if you replace "foo/bar" with "foo" and this happens, only the tests related to the backslash rule would fail).
- Let's use for this test the exact same message that we had in the TODO (because it was a realistic message that could happen in real life; this way it is more understandable why we need this exception).

knocte:
@realmarv you forgot to remove the TODO item :)

commitlintplugins.test.ts

commitlintplugins.test.ts

realmarv:
> @realmarv you forgot to remove the TODO item :)

Done!

knocte:
I've realised we should also allow camelCase (not just PascalCase) in this PR, in case the dev is using a language that defaults to this style (e.g. JavaScript). Let's add a new test covering this and update the rule please.

commitlint.config.ts
```diff
@@ -97,6 +96,16 @@ module.exports = {
...
+
+                        if (subject != null && subject.length > 1) {
+                            let firstWord = subject.trim().split(' ')[0];
+                            if (firstWord[0].toUpperCase() === firstWord[0]){
+                                let numUpper = firstWord.length - firstWord.replace(/[A-Z]/g, '').length;
+                                if (numUpper >= 2){
+                                    offence = false;
+                                }
+                            }
+                        }
```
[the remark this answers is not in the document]
realmarv: Done!

knocte:
Sorry for the constant rebasing request, but it's very difficult to not get conflicts after merging something to master :'( :)

commitlint.config.ts
```diff
@@ -76,6 +76,15 @@ function isFooterNote(line: string): boolean {
         isFixesSentence(line);
 }
 
+function isName(word: string) {
```
knocte, line 79: @realmarv this needs an assert

knocte: @realmarv while you're at it, let's rename the function name to wordIsStartOfSentence

realmarv: > @realmarv while you're at it, let's rename the function name to wordIsStartOfSentence

I didn't get why. Because this function is making sure the first character is uppercase and the rest of the characters are lowercase and there are no non-alphabetical characters in it, which I guess is the definition of a name.

knocte:
> there are no non-alphabetical characters in it

When you said this I realized I may have made a mistake when I answered to a question from you. You asked me if A2 is a name; I said no. But what I meant is that the rule shouldn't care in that case.

realmarv:
> > there are no non-alphabetical characters in it
> 
> When you said this I realized I may have made a mistake when I answered to a question from you. You asked me if A2 is a name; I said no. But what I meant is that the rule shouldn't care in that case.

I'm a bit confused. The rule rejects a Name as a subject's first word but A2 is not a Name so it's ok

knocte:
Thing is, I don't like the function name "isName" because if it's actually a name (let's say, Zeus), then the rule would complain but the developer didn't do any mistake. What the rule is trying to flag here is writing "Increase the foobar" instead of "increase the foobar". And the simplest logic to check this is see if first letter is uppercase while the rest are lowercase (if this is true, then rule should flag). But a real "Name" would be a false positive from the rule. That's why I don't like that name of the function, because it cannot really know if it's name or not (unless it queries Wikipedia? you know what I mean?).

realmarv:
> Thing is, I don't like the function name "isName" because if it's actually a name (let's say, Zeus), then the rule would complain but the developer didn't do any mistake. What the rule is trying to flag here is writing "Increase the foobar" instead of "increase the foobar". And the simplest logic to check this is see if first letter is uppercase while the rest are lowercase (if this is true, then rule should flag). But a real "Name" would be a false positive from the rule. That's why I don't like that name of the function, because it cannot really know if it's name or not (unless it queries Wikipedia? you know what I mean?).

Then the logic of the function is also wrong. Should I ignore it and just rename the function to wordIsStartOfSentence?

knocte:
> Then the logic of the function is also wrong

Please elaborate. And give an example.

knocte:
> As you said, the user might use a real name at the beginning of the subject.

But that problem is not fixable, unless we inject AI.

knocte:
> But that problem is not fixable, unless we inject AI.

That's why I'm proposing the change of the function name, to at least not highlight the limitations of the code.

realmarv:
> > But that problem is not fixable, unless we inject AI.
> 
> That's why I'm proposing the change of the function name, to at least not highlight the limitations of the code.

right

PR #10, opened 2022-10-17
A rule that makes sure the body paragraphs start with an upper case letter and end with a dot

realmarv:


commitlint.config.ts
```diff
@@ -80,6 +128,8 @@ module.exports = {
                     ];
                 },
 
+
+
```
knocte, line 132: @realmarv nit, these 2 lines above don't need to be added

knocte:
Is the title of this PR correct? It says "A rule that makes sure the body starts with an upper case letter and ends with a dot" but I thought it would check this for every paragraph that is in the body, not just once, right?

commitlint.config.ts
realmarv: Done!

realmarv:
> Is the title of this PR correct? It says "A rule that makes sure the body starts with an upper case letter and ends with a dot" but I thought it would check this for every paragraph that is in the body, not just once, right?

Done!

commitlint.config.ts
```diff
@@ -23,6 +23,7 @@ function isBigBlock(line: string) {
 module.exports = {
     parserPreset: 'conventional-changelog-conventionalcommits',
     rules: {
+        'body-prose': [RuleStatus.Error, 'always'],
```
knocte, line 26: @realmarv you put this rule the first, but strictly speaking if following alphabetical order, it should be the second; however I think here we're just following order of execution and I think it's better to place less important rules the last, so let's put this one one of the last ones, like after subject-lowercase

commitlint.config.ts
```diff
@@ -61,6 +62,53 @@ module.exports = {
...
+                            // It's a paragraph that only consists of a block
+                            if (/^```[^]*```$/.test(par.trim())){
+                                continue;
+                            }
+
+                            par = par.replace(/```[^]*```/g, '').trim();
+
+                            let firstIsUpperCase = par[0].toUpperCase() == par[0];
+                            let firstIsLowerCase = par[0].toLowerCase() == par[0];
+                            let startWithLowerCase = firstIsLowerCase && (!firstIsUpperCase);
```
knocte, line 83: @realmarv how about we create a function called IsUpperCase that calls .toUpperCase and .toLowerCase under the hood? this way this code would be clearer

knocte: (and other code in other existing rules)

commitlint.config.ts
```diff
@@ -61,6 +62,53 @@ module.exports = {
...
+                            let endsWithDotOrColon = par[par.length - 1] === '.' || par[par.length - 1] === ':';
+                            
+                            if (startWithLowerCase || !endsWithDotOrColon){
+                                let line = par.split(/\r?\n/)[0];
+                                
+                                // it's a URL
+                                let containsASpace = line.indexOf(" ") >= 0;
+                                
+                                // it's a footer reference, i.e. [1] someUrl://foo/bar/baz
+                                let startsWithRef = (line[0] === "[" && line.indexOf("] ") > 0);
```
knocte, line 94: @realmarv I believe you're copying code from other rules here, right? if so, let's not duplicate code, let's create functions that both rules use please

commitlintplugins.test.ts
```diff
@@ -4,12 +4,85 @@ function runCommitLintOnMsg(inputMsg: string) {
...
+    // because ```blocks surrounded like this``` can bypass the limit
+    expect(bodyProse6.status).toBe(0);
+});
+
+
+test('body-prose7', () => {
+    let commitMsgWithParagraphEndingWithColon =
+        "foo: this is only a title" + "\n\n" + "Bar baz:\n\nBlah blah.";
+    let bodyProse7 = runCommitLintOnMsg(commitMsgWithParagraphEndingWithColon);
+    //console.log("=============>" + trailingWhitespace6.stdout);
```
knocte, line 67: @realmarv please watch out for copy+paste errors, this comment needs changing ^

commitlintplugins.test.ts
```diff
@@ -4,12 +4,85 @@ function runCommitLintOnMsg(inputMsg: string) {
...
+    expect(bodyProse5.status).toBe(0);
+});
+
+
+test('body-prose6', () => {
+    let commitMsgWithBlock =
+        "foo: this is only a title" + "\n\n" + "Bar baz.\n```\n" + 'if (foo) {bar(); }' + "\n```";
+    let bodyProse6 = runCommitLintOnMsg(commitMsgWithBlock);
+
+    // because ```blocks surrounded like this``` can bypass the limit
```
knocte, line 58: @realmarv this comment needs changing too because body-prose is not about limits

commitlint.config.ts
knocte: you wrote "Done!" here (and pressed the button "Resolve conversation") before pushing the commit that resolves this

commitlint.config.ts
realmarv: Done!

commitlint.config.ts
realmarv: Done!

knocte:
PR's CI is not green :) (pls not this CI is not your branch's CI)

commitlint.config.ts
```diff
@@ -61,6 +62,53 @@ module.exports = {
...
         {
             rules: {
+                'body-prose': ({body}: {body:any}) => {
+                    let offence = false;
+
+                    // does msg have a body?
+                    if (body !== null) {
+                        let bodyStr = convertAnyToString(body, "body");
+                        
+                        for (let par of bodyStr.trim().split('\n\n')){
```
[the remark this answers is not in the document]
realmarv: Done!

commitlintplugins.test.ts

commitlint.config.ts
realmarv: Done!

commitlint.config.ts
knocte: this is not resolved yet, I still see the hunk

commitlint.config.ts
```diff
@@ -107,14 +176,7 @@ module.exports = {
...
                         if (subject != null && subject.length > 1) {
-                            let firstIsUpperCase = subject[0].toUpperCase() == subject[0];
-                            let firstIsLowerCase = subject[0].toLowerCase() == subject[0];
-                            let secondIsUpperCase = subject[1].toUpperCase() == subject[1];
-                            let secondIsLowerCase = subject[1].toLowerCase() == subject[1];
-
-                            offence = firstIsUpperCase && (!firstIsLowerCase)
-                                // to whitelist acronyms
-                                && (!secondIsUpperCase) && secondIsLowerCase;
+                            offence = isUpperCase(subject[0]) && isLowerCase(subject[1])
```
knocte, line 179: missing a semicolon at the end of the line

commitlintplugins.test.ts
```diff
@@ -4,12 +4,87 @@ function runCommitLintOnMsg(inputMsg: string) {
...
+    expect(bodyProse7.status).toBe(0);
+});
+
+
+test('body-prose8', () => {
+    let commitMsgWithBlockParagraph=
+        "foo: this is only a title" + "\n\n" + "Bar baz:\n\n```\ntype Foo =\n    string\n```";
+    let bodyProse8 = runCommitLintOnMsg(commitMsgWithBlockParagraph);
+
+    // because paragraphs that only consist of a block can bypass the rule
```
knocte, line 78: let's just say that a block is not a paragraph

knocte: actually, shouldnt this be covered by body-prose6 test? I see this test as redundant

commitlintplugins.test.ts
```diff
@@ -4,12 +4,87 @@ function runCommitLintOnMsg(inputMsg: string) {
...
+    expect(bodyProse3.status).toBe(0);
+});
+
+
+test('body-prose4', () => {
+    let commitMsgWithFootnoteUrl =
+        "foo: this is only a title" + "\n\n" + "Bla blah[1] bla.\n\n[1] someUrl://blahblah.com";
+    let bodyProse4 = runCommitLintOnMsg(commitMsgWithFootnoteUrl);
+
+    // because URLs in footer can bypass the limit
```
knocte, line 38: @realmarv same here, this rule is not a char length limit

commitlint.config.ts
```diff
@@ -167,14 +229,14 @@ module.exports = {
                             if (line.length > bodyMaxLineLength) {
 
                                 // it's a URL
-                                let containsASpace = line.indexOf(" ") >= 0;
+                                let containsSpace = containsASpace(line);
```
knocte, line 232: I'm thinking this would be more readable if the function is called mightBeUrl, and returns the opposite of "contains a space"; this way it's a function that is more aligned with the other two

realmarv: Done!

commitlintplugins.test.ts
realmarv: > let's just say that a block is not a paragraph

Do you mean this is not a valid body?

commitlintplugins.test.ts
```diff
@@ -4,12 +4,87 @@ function runCommitLintOnMsg(inputMsg: string) {
...
+    expect(bodyProse2.status).toBe(0);
+});
+
+
+test('body-prose3', () => {
+    let commitMsgWithUrl =
+        "foo: this is only a title" + "\n\n" + "someUrl://blahblah.com";
+    let bodyProse3 = runCommitLintOnMsg(commitMsgWithUrl);
+
+    // because URLs can bypass the limit
```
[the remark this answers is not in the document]

commitlintplugins.test.ts
```diff
@@ -4,12 +4,87 @@ function runCommitLintOnMsg(inputMsg: string) {
...
+    expect(bodyProse4.status).toBe(0);
+});
+
+
+test('body-prose5', () => {
+    let commitMsgWithBugUrl =
+        "foo: this is only a title" + "\n\n" + "Fixes someUrl://blahblah.com";
+    let bodyProse5 = runCommitLintOnMsg(commitMsgWithBugUrl);
+
+    // because URLs in "Fixes <URL>" sentence can bypass the limit
```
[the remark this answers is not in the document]

commitlint.config.ts

commitlintplugins.test.ts
knocte: I mean to change the comment

knocte: Thing is, I think body-prose6 should have \n\n as separation between first paragraph and the block. If you change it this way, you understand that this test is the same right?

realmarv: Right. Done!

commitlint.config.ts
knocte: @realmarv ping ^

commitlintplugins.test.ts
```diff
@@ -4,12 +4,77 @@ function runCommitLintOnMsg(inputMsg: string) {
...
+    let bodyProse5 = runCommitLintOnMsg(commitMsgWithBugUrl);
+
+    // because URLs in "Fixes <URL>" sentence can bypass the rule
+    expect(bodyProse5.status).toBe(0);
+});
+
+
+test('body-prose6', () => {
+    let commitMsgWithBlock =
+        "foo: this is only a title" + "\n\n" + "Bar baz.\n\n```\n" + 'if (foo) {bar(); }' + "\n```";
```
knocte, line 55: why are you wrapping the code block with single-quotes, and why are you splitting this string so many times? IMO it's simpler to just write:

```
let commitMsgWithBlock =
        "foo: this is only a title\n\nBar baz.\n\n```\nif (foo) { bar(); }\n```";
```

commitlint.config.ts
```diff
@@ -20,6 +20,32 @@ function isBigBlock(line: string) {
     return (line.length == bigBlockDelimiter.length) && (line.indexOf("```") == 0);
 }
 
+function isUpperCase(letter: string) {
```
knocte, line 23: Before I merge this PR I want to raise awareness of this nitpick that might become a potential problem in the future, if some developer makes use of this function without being careful: see how the parameter is called "letter", but the type is "string" instead of "char". Now, it turns out the type `char` doesn't exist on TypeScript so we will need to protect ourselves from this pitfall at runtime instead of at compile time (better than nothing...). So let's do some checks before performing the logic in the function, as in "preconditions". For this function I guess the only precondition is that letter's length is just one character (if it's not, just throw an exception). After you're done adding this precondition, add others to other functions, such as isLowerCase, mightBeUrl (see the param is called "line" in that case, so that's also an assumption), and the other two

commitlint.config.ts
realmarv: I think it might be solved in my new commits because I can't find these lines now

commitlint.config.ts
realmarv: Done

PR #11, opened 2022-10-18
A rule that makes sure the second line of a commit message is always empty

realmarv:


knocte:
As I said in telegram, I realised there was already one rule for this: `body-leading-blank`. We need to investigate why it's not working.

realmarv:
Closed and replaced with https://github.com/realmarv/conventions/commits/useBodyLeadingBlackRule because commitlint already has this rule and we just need to enable it.

PR #13, opened 2022-10-20
New rule that gives replacement suggestions if title isn't short enough

realmarv:


knocte:
@realmarv please rebase PR

abbreviations.ts
```diff
@@ -0,0 +1,153 @@
...
+    "temporary": "tmp",
+    "third": "3rd",
+    "timer": "tmr",
+    "transaction": "tx",
+    "transactions": "txs",
+    "unix to dos": "unix2dos",
+    "value": "val",
+    "variable": "var",
+    "yes/no": "y/n"
+}
```
knocte, line 153: please add EOL at EOF

commitlint.config.ts
```diff
@@ -27,7 +30,7 @@ module.exports = {
         'body-soft-max-line-length': [RuleStatus.Error, 'always'],
         'footer-leading-blank': [RuleStatus.Warning, 'always'],
         'footer-max-line-length': [RuleStatus.Error, 'always', 150],
-        'header-max-length': [RuleStatus.Error, 'always', 50],
+        'header-max-length-with-suggestions': [RuleStatus.Error, 'always'],
```
knocte, line 33: why remove 50 as a parameter? I mean, it's good that you created var headerMaxLineLength, but it would be better if we can pass it as a parameter to the rule config

commitlint.config.ts
```diff
@@ -57,11 +60,35 @@ module.exports = {
...
         // * Detect area(sub-area) in the title that doesn't include area part (e.g., writing (bar) instead of foo(bar))
 
         {
             rules: {
+                'header-max-length-with-suggestions': ({header}: {header:any}) => {
+                    let headerStr = convertAnyToString(header, "header");
+                    let offence = false;
+
+                    let message = `Please do not exceed ${headerMaxLineLength} characters in title.`;
+                    if (headerStr.length > 50) {
```
knocte, line 72: and you're not using headerMaxLineLength here! I guess you forgot

commitlint.config.ts
```diff
@@ -57,11 +60,35 @@ module.exports = {
...
+                        offence = true;
+                        let numRecomendations = 0;
+                        Object.entries(abbr).forEach(([key, value]) => {  
+                            if (headerStr.includes(key.toString())){
+                                if (numRecomendations === 0) {
+                                    message = message + 'We recommend the following replacement(s) in your commit title:\n'
```
knocte, line 78: It's better to not use "we/I" in messages so let's change this to "The following replacement(s) in your commit title are recommended:"

commitlint.config.ts

commitlint.config.ts

commitlint.config.ts

abbreviations.ts
```diff
@@ -0,0 +1,153 @@
...
+    "decimal": "dec",
+    "decrease": "dec",
+    "define": "def",
+    "delete": "drop",
+    "deprecated": "obsolete",
+    "destination": "dest",
+    "developer": "dev",
+    "development": "dev",
+    "difference": "diff",
+    "directory": "dir",
```
knocte, line 55: @realmarv let's add some more: folder -> dir, integer -> int, unsigned integer -> uint, order by -> sort by

commitlint.config.ts
```diff
@@ -57,11 +60,35 @@ module.exports = {
         // * Workflow: detect if wip commit in a branch not named "wip/*" or whose name contains "squashed".
         // * Allow PascalCase word after colon in title (exception to subject-lowercase rule), e.g.: "End2End: TestFixtureSetup refactor"
         // * Detect if commit hash mention in commit msg actually exists in repo.
-        // * Give replacement suggestions in rule that detects too long titles (e.g. and->&, config->cfg, ...)
         // * Detect area(sub-area) in the title that doesn't include area part (e.g., writing (bar) instead of foo(bar))
 
         {
             rules: {
+                'header-max-length-with-suggestions': ({header}: {header:any}, _: any, headerMaxLineLength:any) => {
```
knocte, line 67: @realmarv let's rename this parameter to just maxLineLength (it can be deduced that it is referring to the header, just by looking at the rule name); this way you can uncomment the global var headerMaxLineLength at line 19 and use it at line 30

commitlintplugins.test.ts
```diff
@@ -84,6 +84,46 @@ test('prefer-slash-over-backslash2', () => {
...
+    expect(headerMaxLength1.status).not.toBe(0);
+    expect((headerMaxLength1.stdout + '').includes(expected_message)).toEqual(true)
+});
+
+
+test('header-max-length-with-suggestions2', () => {
+    let commitMsgWithThatExceedsHeaderMaxLength =
+        "foo: this is only a title with a absolute in it that exceeds header max length";
+    let headerMaxLength2 = runCommitLintOnMsg(commitMsgWithThatExceedsHeaderMaxLength);
+    let expected_message = `"absolute" -> "abs"`
```
knocte, line 106: @realmarv instead of testing just another random string, let's modify this test to check that strings with a space also work, e.g. "1 second" -> "1sec"

knocte:
And this needs to be rebased again sorry :(

realmarv:
> And this needs to be rebased again sorry :(

No worries. I understand

commitlintplugins.test.ts

commitlint.config.ts

abbreviations.ts
realmarv: Done

PR #14, opened 2022-10-20
A new rule that checks for too many spaces in the commit message

realmarv:


knocte:
Good work! This commit should also change the last fmt invocation in the recommended command by the body-soft-max-line-length rule to avoid using 2 spaces. See: https://stackoverflow.com/a/14303634/1623521

realmarv:
> Good work! This commit should also change the last fmt invocation in the recommended command by the body-soft-max-line-length rule to avoid using 2 spaces. See: https://stackoverflow.com/a/14303634/1623521

I'm not sure about this StackOverflow discussion because it seems fmt doesn't have -d option. But -u option ensures one space between words and two after sentences. Maybe I should use this one

knocte:
>  it seems fmt doesn't have -d option

What? fmt -d works for me in my macOS laptop.

BTW this PR needs to be rebased.

realmarv:
> > it seems fmt doesn't have -d option
> 
> What? fmt -d works for me in my macOS laptop.
> 
> BTW this PR needs to be rebased.

Maybe the fmt version is different. mine is 8.30

knocte:
> Maybe the fmt version is different. mine is 8.30

How did you check that?

realmarv:
> > Maybe the fmt version is different. mine is 8.30
> 
> How did you check that?

fmt --version

knocte:
> fmt --version command

Interesting, I already tried that, and it doesn't work here. I'll try your -u suggestion and report here.

In the meantime, please fix CI because it's not green.

knocte:
> But -u option ensures one space between words and two after sentences. Maybe I should use this one

Wait, if it uses 2 after sentences, then the new rule created for this PR would make it not pass.

realmarv:
> > But -u option ensures one space between words and two after sentences. Maybe I should use this one
> 
> Wait, if it uses 2 after sentences, then the new rule created for this PR would make it not pass.

How about allowing two spaces (not more than two) after dots?

knocte:
> How about allowing two spaces (not more than two) after dots?

But with the -d argument we avoid that, can you show me the error you get when you use it?

knocte:
> How about allowing two spaces (not more than two) after dots?

Nevermind, you convinced me, let's do this.

realmarv:
> > How about allowing two spaces (not more than two) after dots?
> 
> Nevermind, you convinced me, let's do this.

Done

commitlint.config.ts
```diff
@@ -174,7 +174,7 @@ module.exports = {
 
                 'too-many-spaces': ({raw}: {raw:any}) => {
                     let rawStr = convertAnyToString(raw, "raw");
-                    rawStr = removeCodeBlock(rawStr);
+                    rawStr = removeAllCodeBlocks(rawStr);
```
knocte, line 177: @realmarv there's another rule that uses this technique so let's change it to use the new function (remember, in general, as a programmer, you should never duplicate code)

commitlintplugins.test.ts
```diff
@@ -201,6 +201,47 @@ test('subject-lowercase7', () => {
...
+    let commitMsgWithTwoSpacesAfterSentence =
+        "foo: this is only a title" + "\n\n" + "Bla blah.  blah bla.";
+    let tooManySpaces4 = runCommitLintOnMsg(commitMsgWithTwoSpacesAfterSentence);
+    expect(tooManySpaces4.status).toBe(0);
+});
+
+
+test('too-many-spaces5', () => {
+    let commitMsgWithThreeSpacesAfterSentence =
+        "foo: this is only a title" + "\n\n" + "Bla blah.   blah bla.";
```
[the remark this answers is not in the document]

commitlint.config.ts
realmarv: Done

commitlint.config.ts
```diff
@@ -265,7 +281,7 @@ module.exports = {
 
                     // taken from https://stackoverflow.com/a/66433444/544947 and https://unix.stackexchange.com/a/25208/56844
                     let recommendedUnixCommand =
-                        'git log --format=%B -n 1 $(git log -1 --pretty=format:"%h") | cat - > log.txt ; fmt -w 1111 -s log.txt > ulog.txt && fmt -w 64 -s ulog.txt > wlog.txt && git commit --amend -F wlog.txt';
+                        'git log --format=%B -n 1 $(git log -1 --pretty=format:"%h") | cat - > log.txt ; fmt -w 1111 -s log.txt > ulog.txt && fmt -w 64 -s -u ulog.txt > wlog.txt && git commit --amend -F wlog.txt';
```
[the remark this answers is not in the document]
realmarv: I guess it's about removing extra spaces between the words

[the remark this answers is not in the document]
knocte: in macOS, it already reduces for me any 2 spaces to 1, without your -u addition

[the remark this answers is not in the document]
realmarv: I see. However, in Linux, it doesn't remove extra spaces without the -u option.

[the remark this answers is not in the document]
knocte: You're right! But in macOS it gives an error :( so let's improve the error message so that it shows both the Linux and the macOS command (and while you're at it, let's rebase the PR).

commitlint.config.ts
```diff
@@ -264,12 +280,19 @@ module.exports = {
                     }
 
                     // taken from https://stackoverflow.com/a/66433444/544947 and https://unix.stackexchange.com/a/25208/56844
-                    let recommendedUnixCommand =
-                        'git log --format=%B -n 1 $(git log -1 --pretty=format:"%h") | cat - > log.txt ; fmt -w 1111 -s log.txt > ulog.txt && fmt -w 64 -s ulog.txt > wlog.txt && git commit --amend -F wlog.txt';
+                    function getUnixCommand(fmt_option: string){
```
knocte, line 283: @realmarv we use camelCase for params, not snake_case

commitlint.config.ts
```diff
@@ -264,12 +280,19 @@ module.exports = {
...
+                        fmt_option = fmt_option.trim();
+                        if (fmt_option !== '') { 
+                            fmt_option = ' ' + fmt_option;
```
knocte, line 286: @realmarv why not pass the parameter with a space instead of doing this in the function? the less code we have, the better

realmarv: I thought it might be confusing for someone who is using the function. It's a simple function though

knocte: but you could still add the space in the command string instead of having this logic

commitlint.config.ts

commitlint.config.ts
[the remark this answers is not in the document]

commitlint.config.ts
```diff
@@ -428,6 +432,18 @@ module.exports = {
                     ];
                 },
 
+                'too-many-spaces': ({raw}: {raw:any}) => {
+                    let rawStr = convertAnyToString(raw, "raw");
+                    rawStr = removeAllCodeBlocks(rawStr);
+                    let offence = (rawStr.match(`[^.]  `) !== null);
+                    // && (rawStr.match(`\\.   `) === null)
```
knocte, line 439: @realmarv this line above is a bad practice; I thought about explaining it here to you but I ended up rather explaining it in our WorkflowGuidelines document; please read it there and ask me if you have any questions

knocte: @realmarv and BTW whenever you want to start coding in the branch for this PR again, remember to do git pull first because I added 2 commits myself

knocte:
This needs a rebase because I just merged another PR.

commitlint.config.ts
realmarv: > @realmarv this line above is a bad practice; I thought about explaining it here to you but I ended up rather explaining it in our WorkflowGuidelines document; please read it there and ask me if you have any questions

Do you mean removing those committed lines?

knocte: removing is one way of fixing the bad practice

realmarv: This case `(rawStr.match('\\.   ') === null) `is automatically handled by` (rawStr.match('[^.]  ') !== null)`. Because when there are 3 spaces somewhere the last two spaces will always match this pattern `'[^.]  '` . That's why I commented it. But I'm not sure if I should explain it in a comment or just remove the commented line.

realmarv:
> This needs a rebase because I just merged another PR.

Done

commitlint.config.ts
knocte: > But I'm not sure if I should explain it in a comment or just remove the commented line.

Both options are valid, but leaving the comment there without explanation is the bad practice.

PR #15, opened 2022-10-24
A rule that rejects obvious words in the commit title

realmarv:


commitlintplugins.test.ts
```diff
@@ -87,6 +87,38 @@ test('prefer-slash-over-backslash2', () => {
...
+test('reject-obvious-words3', () => {
+    let commitMsgWithoutObviousWords = "foo: bla bla bla";
+    let rejectObviousWord3 = runCommitLintOnMsg(commitMsgWithoutObviousWords);
+    // console.log('==========>' + rejectObviousWord3.stdout)
+    expect(rejectObviousWord3.status).toBe(0);
+});
+
+
+test('reject-obvious-words4', () => {
+    let commitMsgWithObviousWordsInTheMiddleOfTitle = "foo: bla update bla";
```
knocte, line 115: @realmarv given that we cannot make a rule decide if its failure is a warning or an error, then failing on this case it's too harsh; we need to make the rule only watch out for the word if it's the first word after the area/sub-area.

realmarv:
This PR is closed and replaced with https://github.com/nblockchain/conventions/pull/67 because some commits needed to be squashed and rebased.

PR #16, opened 2022-10-24
CI: switch to vanilla VMs

parhamsaremi:
Install everything ourselves instead of using pre-installed images of github.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,12 +5,21 @@ on: [push, pull_request]
 jobs:
   tests:
     runs-on: ubuntu-20.04
+    container:
+      image: "ubuntu:20.04"
     steps:
       - uses: actions/checkout@v2
       - name: Install required dependencies
         run: |
-          sudo apt install --yes npm
+          apt update
+          apt install -y sudo
+          sudo apt install -y git
+          
+          sudo apt install -y curl
+          curl -sL https://deb.nodesource.com/setup_18.x | sudo -E bash -
+          sudo apt install -y nodejs
           npm install @commitlint/config-conventional
+          sudo npm install -g yarn
           yarn add --dev jest typescript ts-jest @types/jest
       - name: Print versions
         run: |
@@ -23,14 +32,26 @@ jobs:
 
   sanitycheck:
     runs-on: ubuntu-20.04
+    container:
+      image: "ubuntu:20.04"
     steps:
+      - name: Install required dependencies
+        run: |
+          apt update && apt install -y sudo
+          sudo apt install -y git
+          
+          sudo apt install -y curl
+          curl -sL https://deb.nodesource.com/setup_18.x | sudo -E bash -
+          sudo apt install -y nodejs
       - uses: actions/checkout@v2
         with:
           submodules: recursive
           # needed because of commit-lint, see https://github.com/conventional-changelog/commitlint/issues/3376
           fetch-depth: 0
-      - name: Install dependencies of commitlint
-        run: sudo apt install --yes npm && npm install @commitlint/config-conventional
+      # workaround for https://github.com/actions/runner/issues/2033
+      - name: ownership workaround
+        run: git config --global --add safe.directory '*'
+      - name: Install commitlint
+        run: npm install @commitlint/config-conventional
       - name: Validate current commit (last commit) with commitlint
         run: npx commitlint --from HEAD~1 --to HEAD --verbose
-


.github/workflows/ci.yml
```diff
@@ -5,12 +5,23 @@ on: [push, pull_request]
...
     runs-on: ubuntu-20.04
+    container:
+      image: "ubuntu:20.04"
     steps:
       - uses: actions/checkout@v2
       - name: Install required dependencies
         run: |
-          sudo apt install --yes npm
+          apt-get update
+          apt-get install -y sudo
```
knocte, line 15: @parhamsaremi please replace all uses of `apt-get` with `apt`

knocte:
> Co-authored-by: "Andres G. Aragoneses" <knocte@gmail.com>

No need to add this, I only initiated this with adding the "container" tag.

knocte:
> Used curl and nodesource for installing nodejs instead of apt
> becuase of the following error:

Let's mention the version of ubuntu where that error comes from. Also: typo becuase -> because

.github/workflows/ci.yml
```diff
@@ -1,16 +1,23 @@
-name: Run unit tests & check conventions
```
knocte, line 1: @parhamsaremi why are you deleting this?

.github/workflows/ci.yml
```diff
@@ -20,17 +27,28 @@ jobs:
...
-      - name: Install dependencies of commitlint
-        run: sudo apt install --yes npm && npm install @commitlint/config-conventional
+      - name: ownership workaround
+        run: git config --global --add safe.directory '*'
+      - name: Install commitlint
+        run: npm install @commitlint/config-conventional
       - name: Validate current commit (last commit) with commitlint
-        run: npx commitlint --from HEAD~1 --to HEAD --verbose
-
+        run: npx commitlint --from HEAD~1 --to HEAD --verbose
```
knocte, line 54: @parhamsaremi please add EOL before EOF

.github/workflows/ci.yml
```diff
@@ -20,17 +27,28 @@ jobs:
...
+          curl -sL https://deb.nodesource.com/setup_18.x | sudo -E bash -
+          sudo apt-get install -y nodejs
       - uses: actions/checkout@v2
         with:
           submodules: recursive
           # needed because of commit-lint, see https://github.com/conventional-changelog/commitlint/issues/3376
           fetch-depth: 0
-      - name: Install dependencies of commitlint
-        run: sudo apt install --yes npm && npm install @commitlint/config-conventional
+      - name: ownership workaround
```
knocte, line 49: @parhamsaremi in geewallet there's a comment that has a URL directing to the bug; if you don't include it, the next developer will think "workaround to what issue???"

.github/workflows/ci.yml
knocte: this is not done (same problem in the other yml file)

.github/workflows/ci.yml
parhamsaremi: oops this was a mistake.

.github/workflows/ci.yml
parhamsaremi: :+1:

.github/workflows/ci.yml
parhamsaremi: right I guess I missed that :+1:

.github/workflows/ci.yml
parhamsaremi: sorry, it was done when I commented done. but I needed to use an older version and I forgot to re-apply these.

knocte:
Can you rebase this PR please? You'll know why I want it to be rebased as soon as you do it.

.github/workflows/ci.yml
```diff
@@ -5,12 +5,22 @@ on: [push, pull_request]
...
+      image: "ubuntu:20.04"
     steps:
       - uses: actions/checkout@v2
       - name: Install required dependencies
         run: |
-          sudo apt install --yes npm
+          apt update
+          apt install -y sudo
+          sudo apt install -y git
+          sudo apt install -y gcc g++ make
```
knocte, line 17: @parhamsaremi why are gcc and g++ and make needed?

parhamsaremi: They were in the steps of installing nodejs :thinking:. I can remove them and check if there'll be a problem. :thinking:

parhamsaremi:
> Can you rebase this PR please? You'll know why I want it to be rebased as soon as you do it.

Done

knocte:
You cheated haha, why did you change "error" to "Error"? you shouldn't change a message that you didn't write.

parhamsaremi:
> You cheated haha, why did you change "error" to "Error"? you shouldn't change a message that you didn't write.

:laughing: I saw that it failed the CI, so I guessed I should fix it :thinking:

.github/workflows/ci.yml
parhamsaremi: I tested it in this branch: https://github.com/parhamsaremi/conventions/commits/vanillaCI-test-gcc

I guess it's not needed. I'll remove it now.

knocte:
> I saw that it failed the CI, so I guessed I should fix it

What did the CI say when it failed? I'm sure it didn't tell you "please change the error messages of your program to accomodate our rules", right?

parhamsaremi:
> > I saw that it failed the CI, so I guessed I should fix it
> 
> What did the CI say when it failed? I'm sure it didn't tell you "please change the error messages of your program to accomodate our rules", right?

right. it was problem with body rules. Captial case in the beginning and no "." in the end :thinking: . :thinking:

knocte:
> it was problem with body rules.

That is not saying the specific error message that you got. I want you to be specific so that we understand what to do here.

parhamsaremi:
I believe the 

> > it was problem with body rules.
> 
> That is not saying the specific error message that you got. I want you to be specific so that we understand what to do here.

I believe the first to do here is to revert the commit message and push again to see what errors it will generate. should I do this?

knocte:
> I believe the first to do here is to revert the commit message and push again to see what errors it will generate. should I do this?

knocte:
> I believe the first to do here is to revert the commit message and push again to see what errors it will generate. should I do this?

No need, I'm sure you still have access to the log, in the Actions tab.

parhamsaremi:
> > I believe the first to do here is to revert the commit message and push again to see what errors it will generate. should I do this?
> 
> No need, I'm sure you still have access to the log, in the Actions tab.

right, here it is:
```
error jest-cli@29.2.1: The engine "node" is incompatible with
this module. Expected version "^[14](https://github.com/nblockchain/conventions/actions/runs/3319956109/jobs/5485751456#step:4:15).[15](https://github.com/nblockchain/conventions/actions/runs/3319956109/jobs/5485751456#step:4:16).0 || ^16.10.0 ||
>=18.0.0". Got "10.19.0"

error Found incompatible module.
✖   Please begin a paragraph with uppercase letter and end it with a dot [body-prose]

✖   found 1 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
```

knocte:
Great, so I ask you, was your error message a paragraph?

parhamsaremi:
> > Great, so I ask you, was your error message a paragraph?
> > I think this was the complete message, although I'm not sure about the warning part.
> 
> ```
> warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.
> [1/4] Resolving packages...
> [2/4] Fetching packages...
> error jest-cli@29.2.1: The engine "node" is incompatible with this module. Expected version "^14.15.0 || ^16.10.0 || >=18.0.0". Got "10.19.0"
> error Found incompatible module.
> info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
> ```

I just started from the error because I thought the warning is not related and the next two lines seemed unrelated too. :thinking:

knocte:
You didn't answer my question.

parhamsaremi:
> You didn't answer my question.

I guess I didn't get your question the first time :thinking: I guess it's not a paragraph and it's just an error message :thinking:.

knocte:
Exactly, so don't try to convert your error message into a paragraph. You need to make commitlint understand that you're not writing a paragraph. How do you do it here, for example? Let me do it here:

```
warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.
[1/4] Resolving packages...
[2/4] Fetching packages...
error jest-cli@29.2.1: The engine "node" is incompatible with this module. Expected version "^14.15.0 || ^16.10.0 || >=18.0.0". Got "10.19.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
```

What did I just do with the error message?

parhamsaremi:
> Exactly, so don't try to convert your error message into a paragraph. You need to make commitlint understand that you're not writing a paragraph. How do you do it here, for example? Let me do it here:
> 
> ```
> warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.
> [1/4] Resolving packages...
> [2/4] Fetching packages...
> error jest-cli@29.2.1: The engine "node" is incompatible with this module. Expected version "^14.15.0 || ^16.10.0 || >=18.0.0". Got "10.19.0"
> error Found incompatible module.
> info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
> ```
> 
> What did I just do with the error message?

ummm I can't see any difference! I even used diffchecker to check it. can you point it out? :thinking:

knocte:
Man, don't you see it has a grey background? compared to the rest of the text.

knocte:
> so you mean we should add "```" am I right?

Right, if you wrap text with 3 ticks then you're writing special things inside a block, like code, error messages, etc (all in all, not paragraphs).

parhamsaremi:
> > so you mean we should add "```" am I right?
> 
> Right, if you wrap text with 3 ticks then you're writing special things inside a block, like code, error messages, etc (all in all, not paragraphs).

I didn't know they were allowed in the commit message :thinking: . so I guess I"ll fix it now using this.

parhamsaremi:
@knocte I think this is fixed.

parhamsaremi:
> Normally, when you use text like "because of the following error", it's better to end the paragraph with a colon, not with a dot.

fixed.

knocte:
> error jest-cli@29.2.1: The engine "node" is incompatible with
this module. Expected version "^[14](https://github.com/nblockchain/conventions/actions/runs/3319956109/jobs/5485751456#step:4:15).[15](https://github.com/nblockchain/conventions/actions/runs/3319956109/jobs/5485751456#step:4:16).0 || ^16.10.0 ||
>=18.0.0". Got "10.19.0"

Now that I think about it, this error alone shouldn't have stopped us. We could always use an older version of jest that is compatible with our version of Node.

But maybe there was another error with the latest version of commitlint not being compatible with older NodeJS? If this is the case, then indeed we cannot use an older version of Node. Please confirm this is the case.

parhamsaremi:
> > error jest-cli@29.2.1: The engine "node" is incompatible with
> > this module. Expected version "^[14](https://github.com/nblockchain/conventions/actions/runs/3319956109/jobs/5485751456#step:4:15).[15](https://github.com/nblockchain/conventions/actions/runs/3319956109/jobs/5485751456#step:4:16).0 || ^16.10.0 ||
> > =18.0.0". Got "10.19.0"
> 
> Now that I think about it, this error alone shouldn't have stopped us. We could always use an older version of jest that is compatible with our version of Node.
> 
> But maybe there was another error with the latest version of commitlint not being compatible with older NodeJS? If this is the case, then indeed we cannot use an older version of Node. Please confirm this is the case.

By looking at previous CIs, I think most of them actually finished with an error regarding the jest. But I think when we updated to 22.04 the same error appeared for commitlint. I'm now double-checking this by pushing to another branch to check what errors I'll get.

parhamsaremi:
> > > error jest-cli@29.2.1: The engine "node" is incompatible with
> > > this module. Expected version "^[14](https://github.com/nblockchain/conventions/actions/runs/3319956109/jobs/5485751456#step:4:15).[15](https://github.com/nblockchain/conventions/actions/runs/3319956109/jobs/5485751456#step:4:16).0 || ^16.10.0 ||
> > > =18.0.0". Got "10.19.0"
> > 
> > 
> > Now that I think about it, this error alone shouldn't have stopped us. We could always use an older version of jest that is compatible with our version of Node.
> > But maybe there was another error with the latest version of commitlint not being compatible with older NodeJS? If this is the case, then indeed we cannot use an older version of Node. Please confirm this is the case.
> 
> By looking at previous CIs, I think most of them actually finished with an error regarding the jest. But I think when we updated to 22.04 the same error appeared for commitlint. I'm now double-checking this by pushing to another branch to check what errors I'll get.

@knocte this is the error I get when I when I use ubuntu 22.04
```
error @commitlint/config-conventional@17.1.0: The engine "node" is incompatible with this module. Expected version ">=v14". Got "12.22.9"
error Found incompatible module.
```

knocte:
Ok thanks I just pushed https://github.com/nblockchain/conventions/commit/4a688427026212ed292356a859cacdca9a4f5f52 to clarify.

knocte:
@parhamsaremi where did you get the installation method with curl? If I follow the download instructions on nodejs website, I don't reach to that.

parhamsaremi:
> @parhamsaremi where did you get the installation method with curl? If I follow the download instructions on nodejs website, I don't reach to that.

I opened this link: https://nodejs.org/en/download/
After that I moved to this section [Installing Node.js via package manager](https://nodejs.org/en/download/package-manager/)

and finally moved to ubuntu section. I saw this method on other sites too.

knocte:
It's very ironic that when you click on something titled "via package manager" it directs you to https://github.com/nodesource/distributions/blob/master/README.md , which gives instructions that ARE NOT using package managers (a package manager is apt, not curl). What you should have done is keep reading in https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions to find the next section called "Alternatives", where it suggests to install nodejs via snap (this is a package manager, similar to apt).

So let's open a new PR to change the curl invocation with the snap installation.

parhamsaremi:
> It's very ironic that when you click on something titled "via package manager" it directs you to https://github.com/nodesource/distributions/blob/master/README.md , which gives instructions that ARE NOT using package managers (a package manager is apt, not curl). What you should have done is keep reading in https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions to find the next section called "Alternatives", where it suggests to install nodejs via snap (this is a package manager, similar to apt).
> 
> So let's open a new PR to change the curl invocation with the snap installation.

I guess I missed the snap part. It was weird to me that it didn't use package manager as well but I thought there was no other option 🤔

knocte:
Let's forget about it because GitHubCI doesn't seem to be compatible with installing snaps :(

PR #17, opened 2022-10-24
Add a new rule that rejects using #XYZ refs in favour for full URLs

realmarv:


commitlintplugins.test.ts
```diff
@@ -152,6 +152,20 @@ test('prefer-slash-over-backslash2', () => {
...
+});
+
+
+test('reject-hashtag-refs2', () => {
+    let commitMsgWithFullUrl = "foo: blah blah" + '\n\n' + 'Fixes someUrl://blah.blah/158';
+    let rejectHashtagRefs1 = runCommitLintOnMsg(commitMsgWithFullUrl);
+    expect(rejectHashtagRefs1.status).toBe(0);
+});
+
+
```
knocte, line 168: @realmarv you're missing a test that has a #123 inside a 3-tick block.

commitlint.config.ts
```diff
@@ -68,6 +68,10 @@ function includeHashtagRef(text: string) {
     return text.match(`#[0-9]+`) !== null;
 }
 
+function removeCodeBlock(text: string) {
```
knocte, line 71: @realmarv does this function only remove one code block? or does it remove any code blocks? (if the former, then it's wrong; if the latter, then its name is wrong)

realmarv: It's the latter. How about removeCodeBlocks or removeAllCodeBlocks?

knocte: removeAllCodeBlocks :+1:

commitlintplugins.test.ts
realmarv: Done

commitlintplugins.test.ts
```diff
@@ -152,6 +152,28 @@ test('prefer-slash-over-backslash2', () => {
 });
 
 
+test('reject-hashtag-refs1', () => {
+    let commitMsgWithHashtagRef = "foo: fixes #123";
```
knocte, line 156: @realmarv woah, no, let's not have a test that has it in the commit title, because we cannot suggest URLs to be added to the commit title (they would make the title too long for sure)

knocte:
We need to add a test here that the rule should ignore: when #xyz is included in commit title, wrapped by parentheses. Now, you might ask, why should the rule ignore this case? Because that usually happens when merging a PR and then the number after '#' doesn't refer to a github issue but to a github pull request.

knocte:
@realmarv I've realised that a title with "Merge #XYZ blah blah" is valid, so let's forget about this rule inspecting the title, let's make it just bind to body. (And let's rebase this PR.)

realmarv:
> @realmarv I've realised that a title with "Merge #XYZ blah blah" is valid, so let's forget about this rule inspecting the title, let's make it just bind to body. (And let's rebase this PR.)

Done

commitlint.config.ts
```diff
@@ -186,6 +194,25 @@ module.exports = {
                     ];
                 },
 
+                'proper-issue-refs': ({raw}: {raw:any}) => {
```
knocte, line 197: @realmarv you're using a workaround to a bug here ^ (using raw instead of body); let's benefit from your bug filing by adding a comment explaining that this is a workaround, and linking to the bug

commitlint.config.ts
```diff
@@ -186,6 +194,25 @@ module.exports = {
...
+                    let offence = false;
+
+                    let rawStr = convertAnyToString(raw, "raw").trim();
+                    let lineBreakIndex = rawStr.indexOf('\n')
+                    
+                    if (lineBreakIndex >= 0){
+                        let bodyStr = rawStr.substring(lineBreakIndex, )
```
knocte, line 204: @realmarv weird way to call the function, I'm surprised it works! let's remove the comma

commitlint.config.ts
```diff
@@ -76,6 +76,14 @@ function isFooterNote(line: string): boolean {
         isFixesSentence(line);
 }
 
+function includeHashtagRef(text: string) {
```
knocte, line 79: the name of this function suggests that it is procedural (because you're using imperative tense), let's rather replace "include" with "includes" to make it sound more like a func that returns a boolean and doesn't perform any action

commitlint.config.ts
realmarv: Oh, I forgot to add rawStr.length after the comma. I'm surprised too.

commitlint.config.ts
realmarv: Done

commitlint.config.ts

commitlint.config.ts

commitlint.config.ts
```diff
@@ -76,6 +76,14 @@ function isFooterNote(line: string): boolean {
         isFixesSentence(line);
 }
 
+function includesHashtagRef(text: string) {
+    return text.match(`#[0-9]+`) !== null;
+}
+
+function removeAllCodeBlocks(text: string) {
```
knocte, line 83: @realmarv I thought that we had agreed already that this logic is already being used in other rules and then we should avoid code duplication?

commitlint.config.ts
```diff
@@ -76,6 +76,14 @@ function isFooterNote(line: string): boolean {
...
+    return text.replace(/```[^]*```/g, ''); // maybe I should include \n in the pattern
```
knocte, line 84: I'm confused by this comment; why don't you do some testing to figure out if \n should be included or not in the pattern?

realmarv: Good point!

realmarv: I removed the comment because I didn't find a failing case

commitlint.config.ts
realmarv: Sorry. Maybe it was in another branch and I didn't notice this one. Done

knocte:
Unfortunately, this needs another rebase.

realmarv:
> Unfortunately, this needs another rebase.

Done

commitlint.config.ts
```diff
@@ -416,6 +424,26 @@ module.exports = {
                     ];
                 },
 
+                'proper-issue-refs': ({raw}: {raw:any}) => {
+                    // workaround for https://github.com/conventional-changelog/commitlint/issues/3412
```
knocte, line 428: this comment is misplaced; right now it looks like you're referring to the variable "offence", which is very confusing

realmarv: How is it now?

knocte: better 👍

PR #20, opened 2022-10-25
Add rule to check if a footer line is in body

parhamsaremi:
Messages that start with "Fixes" or "[]" should be in the footer this newly added rule will check this behavior and will raise errors if it's not correctly addressed.

### DIFF
--- a/commitlint.config.ts
+++ b/commitlint.config.ts
@@ -104,6 +104,7 @@ module.exports = {
         'empty-wip': [RuleStatus.Error, 'always'],
         'footer-leading-blank': [RuleStatus.Warning, 'always'],
         'footer-max-line-length': [RuleStatus.Error, 'always', 150],
+        'footer-notes-misplacement': [RuleStatus.Error, 'always'],
         'header-max-length-with-suggestions': [RuleStatus.Error, 'always', headerMaxLineLength],
         'subject-full-stop': [RuleStatus.Error, 'never', '.'],
         'type-empty': [RuleStatus.Warning, 'never'],
@@ -210,7 +211,35 @@ module.exports = {
                         message
                     ];
                 },
-                
+
+                'footer-notes-misplacement': ({body}: {body:any}) => {
+                    let offence = false;
+
+                    if (body !== null) {
+                        let bodyStr = convertAnyToString(body, "body");
+
+                        let seenBody = false;
+                        let seenFooter = false;
+                        let lines = bodyStr.split(/\r?\n/);
+                        for (let line of lines) {
+                            if (line.length === 0){
+                                continue;
+                            }
+                            seenBody = seenBody || !isFooterNote(line);
+                            seenFooter = seenFooter || isFooterNote(line);
+                            if (seenFooter && !isFooterNote(line)) {
+                                offence = true;
+                                break;
+                            }
+                            
+                        }
+                    }
+                    return [
+                        !offence,
+                        `Footer messages must be placed after body paragraphs, please move any message that starts with a "[]" or "Fixes" to the end of the commmit message.`
+                    ]
+                },
+
                 'prefer-slash-over-backslash': ({header}: {header:any}) => {
                     let headerStr = convertAnyToString(header, "header");
 
--- a/commitlintplugins.test.ts
+++ b/commitlintplugins.test.ts
@@ -181,6 +181,37 @@ test('empty-wip-3', () => {
 });
 
 
+test('footer-notes-misplacement-1', () => {
+    let commitMsgWithRightFooter = "foo: this is only a title" 
+        + "\n\n"+ "Bla bla blah[1]." 
+        + "\n\n" + "Fixes https://some/issue" 
+        + "\n\n" + "[1] http://foo.bar/baz";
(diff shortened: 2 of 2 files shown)


commitlintplugins.test.ts
```diff
@@ -181,6 +181,38 @@ test('empty-wip-3', () => {
...
+})
+
+
+test('footer-correct-place-2', () => {
+    let commitMsgWithWrongFooter = "foo: this is only a title" 
+        + "\n\n" + "Fixes https://some/issue" 
+        + "\n\n"+ "Bla bla blah[1]." 
+        + "\n\n" + "[1] http://foo.bar/baz";
+    let footerCorrectPlace2 = runCommitLintOnMsg(commitMsgWithWrongFooter);
+    console.log('HERE ============>' + footerCorrectPlace2.stdout)
```
knocte, line 200: @parhamsaremi please remove logging (or comment it)

commitlint.config.ts
```diff
@@ -167,6 +173,34 @@ module.exports = {
                     ]
                 },
 
+                'footer-correct-place': ({body}: {body:any}) => {
```
knocte, line 176: @parhamsaremi I'm not convinced about the name of the rule, let's rather name it 'footer-notes-misplacement' (and as a consequence let's rename some vars in the tests)

commitlint.config.ts
```diff
@@ -167,6 +173,34 @@ module.exports = {
...
+                    let offence = false;
+
+                    if (body !== null) {
+                        let bodyStr = convertAnyToString(body, "body");
+
+                        let seenBody = false;
+                        let seenFooter = false;
+                        let lines = bodyStr.split(/\r?\n/);
+                        for (let line of lines) {
+                            if (isEmptyLine(line)){
```
knocte, line 186: @parhamsaremi do we really need this check? let's think about it, would an empty line return true for the function isFooterNote? I guess not, so I don't think we need this extra code (maybe I'm wrong, please double check)

knocte:
@parhamsaremi I think you learned a bad habit from our propensity to ask for squashed branches in geewallet/LN work. Squashing branches is only needed when there are WIP commits (and only if there's more than 1 WIP commit in certain branch). In this case, you worked on this PR for only one day so I believe there wouldn't be any WIP commit. What I was expecting is to see 1 commit first where you add the tests (and CI failing because of that), then a commit where you add the rule and makes the test pass. Please apply TDD as much as you can (and that includes: when using with git).

commitlint.config.ts
```diff
@@ -167,6 +173,34 @@ module.exports = {
...
+                            if (seenFooter && !isFooterNote(line)) {
+                                offence = true;
+                                break;
+                            }
+                            
+                        }
+                    }
+                    return [
+                        !offence,
+                        `Please add footer messages after body paragraphs`
```
knocte, line 200: @parhamsaremi this error message starting with "Please add" would make a developer think that he has to "add stuff" (as in, make the commit message bigger), but this is not what this rule should advice, the thing to be done is to shuffle the content around, so let's not use the word "add" here, it's confusing

knocte (changes requested):

commitlint.config.ts
parhamsaremi: Yes. an empty line won't be considered as footer but !isFooterNote will be true which is not the behavior I want so I skipped empty lines :thinking: (I also commented out this check and one test failed)

parhamsaremi:
> @parhamsaremi I think you learned a bad habit from our propensity to ask for squashed branches in geewallet/LN work. Squashing branches is only needed when there are WIP commits (and only if there's more than 1 WIP commit in certain branch). In this case, you worked on this PR for only one day so I believe there wouldn't be any WIP commit. What I was expecting is to see 1 commit first where you add the tests (and CI failing because of that), then a commit where you add the rule and makes the test pass. Please apply TDD as much as you can (and that includes: when using with git).

Based on our talk last night I learned what I did wrong and I'll work with this standard in future works. Should I split my squashed commit into two new commits in this PR? :thinking:

commitlint.config.ts
knocte: ah fair enough, my bad

knocte: but creating a function for checking if .length is zero looks like overengineering to me :)

knocte: you're not using that function anywhere else... just check if length is zero right there

knocte:
> Should I split my squashed commit into two new commits in this PR?

No need, that was a comment to take in account for future, not an actionable comment. That being said, this PR needs to be rebased sorry.

parhamsaremi:
> > Should I split my squashed commit into two new commits in this PR?
> 
> No need, that was a comment to take in account for future, not an actionable comment. That being said, this PR needs to be rebased sorry.

Yup saw it after addressed the comments and pushed again. it should be ok now.

commitlint.config.ts
parhamsaremi: hmmm sure I'll fix it right now.

parhamsaremi:
@knocte It needed another rebase. now everything should be ok :thinking:

PR #21, opened 2022-10-27
A new rule that suggests using commit hash instead of commit full URL 

realmarv:


commitlintplugins.test.ts
```diff
@@ -160,6 +160,24 @@ test('body-max-line-length7', () => {
...
+        "https://github.com/nblockchain/conventions/commit/3ee07243edc30604088a4b04ca525204ea440710";
+    let commitHashAlone1 = runCommitLintOnMsg(commitMsgWithCommitUrl);
+    expect(commitHashAlone1.status).not.toBe(0);
+});
+
+
+test('commit-hash-alone2', () => {
+    let commitMsgWithCommitHash = 
+        "foo: this is only a title" + "\n\n" + 
+        "This is refering to [1] commit hash.\n\n[1] 3ee07243edc30604088a4b04ca525204ea440710";
```
knocte, line 175: 2 things here:

- typo "refering" -> "referring"
- if you're using [1], then it should be there twice, first where it's used, and at the end as a footer note (if you don't understand this, ask Parham because he recently fixed a unit test that had the same problem)

knocte: oops! nevermind, I didn't see the second [1], scratch that

commitlint.config.ts
```diff
@@ -177,6 +204,28 @@ module.exports = {
...
+                'commit-hash-alone': ({raw}: {raw:any}) => {
+                    let rawStr = convertAnyToString(raw, "raw");
+
+                    let offence = false;
+
+                    let urls = findUrls(rawStr)
+
+                    if (urls !== null) {
+                        for (let url of urls.entries()) {
+                            if (isCommitUrl(url.toString())) {
```
knocte, line 216: @realmarv there's an issue here, we need to detect if the URL belongs to the same repo (if it doesn't, it's fair enough that it is a URL); for this, we would check either git remotes or GitHubAcrtions environment variables

realmarv: Done. Please check [this PR](https://github.com/nblockchain/conventions/pull/23)

realmarv:
This PR is closed and replaced with another [PR](https://github.com/nblockchain/conventions/pull/23). The reason for this replacement is that some of the commits needed to be squashed.

PR #22, opened 2022-10-28
Rule to check every reference is used in body and footer

parhamsaremi:
Prohibit contributors from defining a reference and not putting it in the footer or putting a reference in the footer which is not defined in the body.

### DIFF
--- a/commitlint.config.ts
+++ b/commitlint.config.ts
@@ -262,6 +262,7 @@ module.exports = {
         'footer-leading-blank': [RuleStatus.Warning, 'always'],
         'footer-max-line-length': [RuleStatus.Error, 'always', 150],
         'footer-notes-misplacement': [RuleStatus.Error, 'always'],
+        'footer-references-existence': [RuleStatus.Error, 'always'],
         'header-max-length-with-suggestions': [RuleStatus.Error, 'always', headerMaxLineLength],
         'subject-full-stop': [RuleStatus.Error, 'never', '.'],
         'type-empty': [RuleStatus.Warning, 'never'],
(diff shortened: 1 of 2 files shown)


commitlintplugins.test.ts
```diff
@@ -212,6 +212,34 @@ test('footer-notes-misplacement-3', () => {
...
+        + "\n\n"+ "Bla bla blah."
+        + "\n\n" + "[1] http://foo.bar/baz"
+    let footerReferenceExistence2 = runCommitLintOnMsg(commmitMsgwithWrongFooter);
+    expect(footerReferenceExistence2.status).not.toBe(0);
+})
+
+
+test('footer-references-existence3', () => {
+    let commmitMsgwithWrongFooter = "foo: this is only a title"
+        + "\n\n"+ "Bla bla blah[1], and after that [2], then [3]."
```
knocte, line 235: @parhamsaremi let's add one more test, in which [4] is used, but there's no footer note for it.

knocte:
@parhamsaremi why not squash the 2nd commit with the 1st?

commitlintplugins.test.ts
parhamsaremi: I didn't get this. don't we already have [2] which is used in body but there is no footer for it? :thinking:

parhamsaremi:
> @parhamsaremi why not squash the 2nd commit with the 1st?

sure :+1:

commitlintplugins.test.ts
knocte: oh true, ok, then let's change this test, I prefer the number not being used being the last one, not the one in the middle

commitlint.config.ts
```diff
@@ -397,6 +398,47 @@ module.exports = {
...
+                        let bodyStr = convertAnyToString(body, "body");
+                        let lines = bodyStr.split(/\r?\n/);
+                        let bodyReferences = new Set();
+                        let references = new Set();
+                        for (let line of lines) {
+                            let matches = line.match(/(?<=\[)([0-9]+)(?=\])/g);
+                            if (matches === null) {
+                                continue;
+                            }
+                            for (let m of matches){
```
knocte, line 414: let's rename m

parhamsaremi: done :+1:

knocte:
You're missing an area in each commit message (it's easy to find which area you should use just by looking at the history of the commits in this repo). Don't forget next time please.

parhamsaremi:
> You're missing an area in each commit message (it's easy to find which area you should use just by looking at the history of the commits in this repo). Don't forget next time please.

Oh sorry. Ok sure, I'll keep this in mind.

PR #23, opened 2022-10-28
A new rule that suggests using commit hash instead of commit full URL

realmarv:


knocte:
CI is broken and this needs a rebase.

knocte:
This looks ready to merge? except that some console.log calls need to be removed.

PR #24, opened 2022-10-31
A F# script that detects files that don't have EOL at EOF

realmarv:


eofconvention.fsx
```diff
@@ -0,0 +1,23 @@
+open System.IO
+
+let EOFIsEOL(path: string) = 
+    try
+        use sr = new StreamReader (path)
+        let filetext = sr.ReadToEnd()
+        (filetext |> Seq.last = '\n')
+        
+    with ex ->
```
knocte, line 9: @realmarv when will this happen?

eofconvention.fsx
```diff
@@ -0,0 +1,23 @@
...
+        // printfn "%A" ex
+        true
+
+let EOFIsNotEOL(path: string) = not (EOFIsEOL path)
+
+let NotPrefix (value: string) (str: string) =
+    not (str.StartsWith(value))
+
+printfn "%A" (
+    Directory.GetFiles("./", "*.*", SearchOption.AllDirectories)
```
knocte, line 19: I think you don't need the slash, it should work with just "."

eofconvention.fsx
```diff
@@ -0,0 +1,23 @@
+open System.IO
+
+let EOFIsEOL(path: string) =
```
knocte, line 3: @realmarv nit, this is a bad practice called "Primitive Obsession" (look it up for more info); to fix it you should use the type FileInfo instead of string

knocte:
@realmarv let's modify CI to include a call to this script and fail CI if there's any offending file.

eofconvention.fsx
realmarv: I guess when the file is empty

eofconvention.fsx
```diff
@@ -0,0 +1,23 @@
...
+    with ex ->
+        // printfn "%A" ex
+        true
+
+let EOFIsNotEOL(path: string) = not (EOFIsEOL path)
+
+let NotPrefix (value: string) (str: string) =
+    not (str.StartsWith(value))
+
+printfn "%A" (
```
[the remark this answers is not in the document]
realmarv: Should I throw an error? I thought it would be good to mention file names in the error message.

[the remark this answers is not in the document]
knocte: what do you mean exactly with "throw an error"? in F#, erros are exceptions; but exiting with an exitCode different than 0 is also emitting an error (but yeah, printing the offending files first before exiting has to be done)

eofconvention.fsx
knocte: what exception type in particular? just putting a generic exception like this is a bad practice, use the specific type of exception that you got when you tested this

eofconvention.fsx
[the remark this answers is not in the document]
knocte: that will certainly stop everything and give a non-zero exitCode, yes

eofconvention.fsx
realmarv: Done

eofconvention.fsx
realmarv: Done

eofconvention.fsx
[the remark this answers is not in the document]
realmarv: Done

eofconvention.fsx
realmarv: Done

knocte:
@realmarv I see that your CI is hanging because it's asking for user input:

```
debconf: falling back to frontend: Readline
Configuring tzdata
------------------

Please select the geographic area in which you live. Subsequent configuration
questions will narrow this down by presenting a list of cities, representing
the time zones in which they are located.

  1. Africa      4. Australia  7. Atlantic  10. Pacific  13. Etc
  2. America     5. Arctic     8. Europe    11. SystemV
  3. Antarctica  6. Asia       9. Indian    12. US
Geographic area: 
```

The way to workaround/fix this is to attach something in front of the `apt install` command that is causing trouble, like we do in geewallet: https://github.com/nblockchain/geewallet/blob/master/.github/workflows/CI.yml#L124

eofconvention.fsx
```diff
@@ -0,0 +1,29 @@
+open System.IO
+
+let EOFIsEOL(path: FileInfo) = 
+    try
+        use sr = new StreamReader (path.FullName)
+        let filetext = sr.ReadToEnd()
+        (filetext |> Seq.last = '\n')
+        
+    with
+        | :? System.ArgumentException -> printfn "%A" path; true
```
knocte, line 10: @realmarv ok, getting there, now the next thing to do here is identify which line exactly of those 3 lines inside the `try` block is the one causing ArgumentException, and only put 1 line inside it instead of 3

eofconvention.fsx
```diff
@@ -0,0 +1,29 @@
+open System.IO
+
+let EOFIsEOL(path: FileInfo) =
```
knocte, line 3: @realmarv I'd prefer a better name for this function, such as EolAtEof

eofconvention.fsx
```diff
@@ -0,0 +1,29 @@
...
+    try
+        use sr = new StreamReader (path.FullName)
+        let filetext = sr.ReadToEnd()
+        (filetext |> Seq.last = '\n')
+        
+    with
+        | :? System.ArgumentException -> printfn "%A" path; true
+        
+
+let EOFIsNotEOL(path: FileInfo) = not (EOFIsEOL path)
```
knocte, line 13: @realmarv no need to create this function, you can do this in the line that calls "filter" function (less code this way)

eofconvention.fsx
```diff
@@ -0,0 +1,29 @@
...
+    Directory.GetFiles(".", "*.*", SearchOption.AllDirectories) 
+    |> Array.filter (NotPrefix "./node_modules")
+    |> Array.filter (NotPrefix "./.git")
+    |> Array.map (fun x -> FileInfo(x))
+    |> Array.filter EOFIsNotEOL
+
+if invalidFiles.Length > 0 then
+    let mutable message = "The following files doesn't end with EOL:\n" 
+    invalidFiles |> Array.iter (fun x -> message <- message + x.FullName + "\n")
+    raise (System.Exception(message))
```
knocte, line 29: @realmarv nit: a more visually-appealing way to write this with less parentheses is `raise <| System.Exception message`; but you could even learn here the function `failwith`, which is used when you don't want to raise a specific type of exception, but just the generic one, so `failwith message`

eofconvention.fsx
```diff
@@ -0,0 +1,29 @@
...
+let NotPrefix (value: string) (str: string) =
+    not (str.StartsWith(value))
+
+
+let invalidFiles = 
+    Directory.GetFiles(".", "*.*", SearchOption.AllDirectories) 
+    |> Array.filter (NotPrefix "./node_modules")
+    |> Array.filter (NotPrefix "./.git")
+    |> Array.map (fun x -> FileInfo(x))
+    |> Array.filter EOFIsNotEOL
```
knocte, line 24: @realmarv I think you can use Seq. functions here instead of Array. ones; this way, if in the future someone changes line 20 to return a list instead of an array, then you wouldn't need the cascading changes (because Seq. functions work both on lists and also arrays)

eofconvention.fsx
```diff
@@ -0,0 +1,29 @@
...
+        let filetext = sr.ReadToEnd()
+        (filetext |> Seq.last = '\n')
+        
+    with
+        | :? System.ArgumentException -> printfn "%A" path; true
+        
+
+let EOFIsNotEOL(path: FileInfo) = not (EOFIsEOL path)
+
+let NotPrefix (value: string) (str: string) =
```
knocte, line 15: @realmarv the parameter names here are not the best; let's give them meaningful names

eofconvention.fsx
```diff
@@ -0,0 +1,29 @@
+open System.IO
+
+let EOFIsEOL(path: FileInfo) = 
+    try
+        use sr = new StreamReader (path.FullName)
```
knocte, line 5: @realmarv sr is not in our list of blessed abbreviations :-P you can use `streamReader` here, unless there's an ogre sitting next to you charging you 1 penny for every key stroke

eofconvention.fsx
```diff
@@ -0,0 +1,29 @@
...
+
+let NotPrefix (value: string) (str: string) =
+    not (str.StartsWith(value))
+
+
+let invalidFiles = 
+    Directory.GetFiles(".", "*.*", SearchOption.AllDirectories) 
+    |> Array.filter (NotPrefix "./node_modules")
+    |> Array.filter (NotPrefix "./.git")
+    |> Array.map (fun x -> FileInfo(x))
```
knocte, line 23: @realmarv nits about this line:

- No need for parentheses when calling FileInfo
- Please use a longer name for variable x (otherwise when we adopt FSharpLint in this repo, it will fail here)

jest.config.js
```diff
@@ -2,4 +2,4 @@
 module.exports = {
   preset: 'ts-jest',
   testEnvironment: 'node',
-};
\ No newline at end of file
+};
```
knocte, line 5: can you undo this change? I wanna see CI fail :)

.github/workflows/ci.yml
```diff
@@ -61,3 +61,7 @@ jobs:
         run: npm install @commitlint/config-conventional
       - name: Validate current commit (last commit) with commitlint
         run: npx commitlint --from HEAD~1 --to HEAD --verbose
+      - name: Install dotnet sdk
+        run: sudo apt install -y dotnet6
+      - name: Run script to make sure all files end with EOL 
+        run: dotnet fsi eofconvention.fsx
```
knocte, line 67: we use camelCase for script names, so can you rename eofconvention to eofConvention?

eofconvention.fsx
```diff
@@ -0,0 +1,26 @@
...
+
+    use streamReader = new StreamReader (path.FullName)
+    let filetext = streamReader.ReadToEnd()
+    
+    if filetext <> "" then
+        (filetext |> Seq.last = '\n')
+    else
+        true
+
+let NotPrefix (prefix: string) (mainStr: string) =
```
knocte, line 13: mainStr still a confusing name, maybe use the same as you have used in line 20?

eofconvention.fsx
```diff
@@ -0,0 +1,26 @@
...
+    not (mainStr.StartsWith(prefix))
+
+let invalidFiles = 
+    Directory.GetFiles(".", "*.*", SearchOption.AllDirectories) 
+    |> Seq.filter (NotPrefix "./node_modules")
+    |> Seq.filter (NotPrefix "./.git")
+    |> Seq.map (fun pathStr -> FileInfo pathStr)
+    |> Seq.filter (fun fileInfo -> not (EolAtEof fileInfo))
+
+if invalidFiles |> Seq.length > 0 then
```
knocte, line 23: The pipe |> operator is good for what you did in lines 16-21, a chain-like thing, but when we're talking about a single function with a single argument, it's just much shorter and readable to not use it

realmarv: What about line 25? Should I change that too?

eofconvention.fsx
realmarv: Done

.github/workflows/ci.yml

eofconvention.fsx
realmarv: Done

eofconvention.fsx
realmarv: Done

eofconvention.fsx
realmarv: Done

eofconvention.fsx
realmarv: Done

eofconvention.fsx
realmarv: Done

eofconvention.fsx
realmarv: Done

eofconvention.fsx
realmarv: Done

eofconvention.fsx
realmarv: I removed the try with block if it's ok. replaced it with if else

jest.config.js
realmarv: Sure. Done

eofconvention.fsx
knocte: well line 25 deserves a different conversation :) because using a mutable variable there can be avoided

eofconvention.fsx
knocte: nice, much better

jest.config.js
knocte: great, good stuff, you can put it back to fix CI

eofconvention.fsx
realmarv: I tried to fix it. But now it looks a bit strange

eofConvention.fsx
```diff
@@ -0,0 +1,30 @@
+open System.IO
+
+let EolAtEof(path: FileInfo) = 
+
+    use streamReader = new StreamReader (path.FullName)
+    let filetext = streamReader.ReadToEnd()
+    
+    if filetext <> "" then
```
knocte, line 8: @realmarv nit: use String.Empty instead of `""`

eofConvention.fsx
```diff
@@ -0,0 +1,30 @@
...
+        (filetext |> Seq.last = '\n')
+    else
+        true
+
+let NotPrefix (prefix: string) (pathStr: string) =
+    not (pathStr.StartsWith(prefix))
```
knocte, line 14: @realmarv no need for parenteses in the StartsWith call

eofConvention.fsx
```diff
@@ -0,0 +1,30 @@
...
+
+let invalidFiles = 
+    Directory.GetFiles(".", "*.*", SearchOption.AllDirectories) 
+    |> Seq.filter (NotPrefix "./node_modules")
+    |> Seq.filter (NotPrefix "./.git")
+    |> Seq.map (fun pathStr -> FileInfo pathStr)
```
knocte, line 20: @realmarv if you move Seq.map to be the first call after Directory.GetFiles then you do less Primitive Obsession

eofConvention.fsx
```diff
@@ -0,0 +1,30 @@
...
+    |> Seq.filter (fun fileInfo -> not (EolAtEof fileInfo))
+
+if Seq.length invalidFiles > 0 then
+    let message = 
+        "The following files doesn't end with EOL:\n" + 
+        (invalidFiles 
+        |> Seq.map (fun x -> x.FullName)
```
knocte, line 27: @realmarv longer variable name

eofConvention.fsx
```diff
@@ -0,0 +1,30 @@
+open System.IO
+
+let EolAtEof(path: FileInfo) =
```
knocte, line 3: @realmarv it's not a string anymore so param should be not called path

knocte:
Last but not least: delete the EOL line (TODO) in ReadMe please.

knocte:
Why did you rebase if there were no conflicts?

realmarv:
> Why did you rebase if there were no conflicts?

Just to clean the commit history by squashing WIP commits. Shouldn't I do it?

knocte:
> Just to clean the commit history by squashing WIP commits. Shouldn't I do it?

Sorry I thought you were rebasing, not squashing. Rebasing is only needed when there are conflicts.

eofConvention.fsx
```diff
@@ -0,0 +1,30 @@
...
+let invalidFiles = 
+    Directory.GetFiles(".", "*.*", SearchOption.AllDirectories) 
+    |> Seq.filter (NotPrefix "./node_modules")
+    |> Seq.filter (NotPrefix "./.git")
+    |> Seq.map (fun pathStr -> FileInfo pathStr)
+    |> Seq.filter (fun fileInfo -> not (EolAtEof fileInfo))
+
+if Seq.length invalidFiles > 0 then
+    let message = 
+        "The following files doesn't end with EOL:\n" +
```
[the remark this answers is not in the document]
realmarv: Done

realmarv:
> Last but not least: delete the EOL line (TODO) in ReadMe please.

Done

eofConvention.fsx
realmarv: Done

eofConvention.fsx
realmarv: Done

eofConvention.fsx
realmarv: Done

jest.config.js
realmarv: Done

eofConvention.fsx
realmarv: Done

PR #25, opened 2022-11-03
Fix body prose false negative when there is a commit reference at the end of paragraph

realmarv:


knocte:
I disagree about the way you implemented this, it should not whitelist commit reference, it should whitelist any URL.

realmarv:
> I disagree about the way you implemented this, it should not whitelist commit reference, it should whitelist any URL.

First I thought I should remove the lines without any spaces but then I notices that the last line of paragraph might only consist of a single word and a dot. 
The other way to search for urls is to find the lines that contain slash and no spaces. What do you think?

knocte:
> then I notices that the last line of paragraph might only consist of a single word and a dot.

Yeah I thought about that too, but it's a very corner case.

> The other way to search for urls is to find the lines that contain slash and no spaces. The other way is to remove commit urls and http(s) urls.

Why complicate things so much? We have now a `isValidUrl` function introduced by one of your PRs. You could reuse it and whitelist any line that doesn't have spaces and is a URL.

PR #27, opened 2022-11-07
A new feature that detects files with mixed line ending

realmarv:
Detecting files with mixed line ending.

### DIFF
--- a/mixedLineEnding.fsx
+++ b/mixedLineEnding.fsx
@@ -1,7 +1,7 @@
 open System.IO
 open System
 
-let MixedLineEnding(fileInfo: FileInfo) = 
+let MixedLineEndings(fileInfo: FileInfo) = 
     use streamReader = new StreamReader (fileInfo.FullName)
     let fileText = streamReader.ReadToEnd()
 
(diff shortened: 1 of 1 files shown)


mixedLineEnding.fsx
```diff
@@ -0,0 +1,32 @@
+open System.IO
+open System
+
+let mixedLineEnding(fileInfo: FileInfo) =
```
knocte, line 4: @realmarv we use PascalCase for functions

realmarv: Done

mixedLineEnding.fsx
```diff
@@ -1,7 +1,7 @@
 open System.IO
 open System
 
-let mixedLineEnding(fileInfo: FileInfo) = 
+let MixedLineEnding(fileInfo: FileInfo) =
```
knocte, line 4: @realmarv oh, and it should be plural, right? because I don't think it's checking for 1 line ending

realmarv: Is using ReadToEnd a bad practice here as well? I guess it doesn't include line endings in each line when I use ReadLines(), right?

knocte: yes, there's no reason to use ReadToLine if it can work with ReadLines()

knocte: > I guess it doesn't include line endings in each line when I use ReadLines(), right?

Oh good point, maybe we cannot use ReadLines() :'(

PR #28, opened 2022-11-07
Fix body prose false negative when there is a URL at the end of paragraph

realmarv:


knocte:
Merged manually here: https://github.com/nblockchain/conventions/commit/3875b828ef692f496811f4e69d8d752e3881361c , good work!

PR #31, opened 2022-11-14

realmarv:


knocte:
CI is red.

commitlintplugins.test.ts
```diff
@@ -348,11 +348,28 @@ test('proper-issue-refs3', () => {
...
+        'This reverts commit 0272f587c7eece147e8d1756116b0b43e11c34ac.';
+    let properRevertMessage1 = runCommitLintOnMsg(commitMsgWithoutProperRevertMessage);
+    expect(properRevertMessage1.status).not.toBe(0);
+});
+
+
+test('proper-revert-message2', () => {
+    let commitMsgWithProperRevertMessage = 
+        'Revert 9bb557e54830690b8a8e403d1b74780d86b07b4c\n\n' + 
+        'We need to revert this commit, because/otherwise bla bla.'
```
knocte, line 367: let's make the commit message identical to the one in the first test, except for adding the reasoning later; we don't need the user to re-write the revert commit message, we want the user to **expand** on it

commitlint.config.ts
```diff
@@ -558,6 +560,30 @@ module.exports = {
...
+    
+                'proper-revert-message': ({raw}: {raw:any}) => {
+                    let offence = false;
+                    
+                    let rawStr = convertAnyToString(raw, "raw").trim();
+                    let lineBreakIndex = rawStr.indexOf('\n')
+                    
+                    if (lineBreakIndex >= 0){
+                        // Extracting bodyStr from rawStr rather than using body directly is a 
+                        // workaround for https://github.com/conventional-changelog/commitlint/issues/3412
```
knocte, line 572: why do we need the workaround here if the tests don't involve references?

knocte:
I'd like to see the commit "set defaultIgnores to false" be green after a red one (so that it's clear that it's what makes the test pass), but I can't see that:

Screenshot 2022-11-15 at 12 58 33 PM

commitlint.config.ts
realmarv: You are right, but shouldn't I add a test that involves references? I thought it was possible in revert messages too

commitlintplugins.test.ts
realmarv: > I'd like to see the commit "set defaultIgnores to false" be green after a red one (so that it's clear that it's what makes the test pass), but I can't see that:
> 
> Screenshot 2022-11-15 at 12 58 33 PM

Oh, you are right. It's a squashed branch, and the previous branch was far behind the master and had conflicts with the master branch so I had to rebase it. Therefore it can't be seen now. What should I do?

realmarv:
This PR is closed and replaced by [another PR](https://github.com/nblockchain/conventions/pull/32) because of some version control problems.

PR #32, opened 2022-11-15

realmarv:


knocte:
@realmarv CI is red

commitlintplugins.test.ts
```diff
@@ -348,11 +348,36 @@ test('proper-issue-refs3', () => {
...
+        'Revert "add abbreviations.ts"\n\n' + 
+        'This reverts commit 0272f587c7eece147e8d1756116b0b43e11c34ac\n' +
+        'because/otherwise bla bla.'
+    let properRevertMessage2 = runCommitLintOnMsg(commitMsgWithProperRevertMessage);
+    expect(properRevertMessage2.status).toBe(0);
+});
+
+
+test('proper-revert-message3', () => {
+    let commitMsgWithProperRevertMessage = 'Revert "add abbreviations.ts"';
```
knocte, line 375: I said that it needs to contain the whole message from the previous test, to expand on it; I said message, not title

knocte: ah damn I'm reading test3 not test2, sorry, let me read test2

commitlint.config.ts
```diff
@@ -558,7 +560,27 @@ module.exports = {
                         `Please use full URLs instead of #XYZ refs.`
                     ];
                 },
+    
+                'proper-revert-message': ({header, body}: {header: any, body: any}) => {
+                    let offence = false;
+
+                    // does msg have a body?
+                    if (header.match(/^[Rr]evert/) !== null) {
```
knocte, line 568: this is wrong, when you do `git revert`, the commit message starts with uppercase, not lowercase

commitlint.config.ts
```diff
@@ -558,7 +560,27 @@ module.exports = {
...
+                        if (body !== null) {
+                            let bodyStr = convertAnyToString(body, "body").trim();
+                            let lines = bodyStr.split('\n');
+                            offence = lines[0].match('This reverts commit ') !== null && lines.length == 1;
+                        }
+                        else {
+                            offence = true;
```
knocte, line 575: this is wrong, if there's no body, it's not a default revert commit

knocte:
Ignore all my comments above, let's rebase this PR first.

commitlintplugins.test.ts
```diff
@@ -348,11 +348,36 @@ test('proper-issue-refs3', () => {
...
+    let properRevertMessage1 = runCommitLintOnMsg(commitMsgWithoutProperRevertMessage);
+    expect(properRevertMessage1.status).not.toBe(0);
+});
+
+
+test('proper-revert-message2', () => {
+    let commitMsgWithProperRevertMessage = 
+        'Revert "add abbreviations.ts"\n\n' + 
+        'This reverts commit 0272f587c7eece147e8d1756116b0b43e11c34ac\n' +
+        'because/otherwise bla bla.'
```
knocte, line 368: @realmarv let's add a test4 that explains the reason in same line where the text "This reverts commit" is

realmarv: That line's length is already 60 characters. If I do that, the test will fail because of the body-max-length rule. unless I say "This reverts commit bellow because bla bla" and put the commit hash in the next line. What do you think?

knocte: I knew you would say that, but I said this because I think your logic is wrong, because it's only looking at the first line. And you're doing that because you know our limit is 64, but what if we increase the limit in the future? adding the reason in the first line could fit in this case, and your rule would have a bug.

commitlint.config.ts
```diff
@@ -558,7 +560,27 @@ module.exports = {
                         `Please use full URLs instead of #XYZ refs.`
                     ];
                 },
+    
+                'proper-revert-message': ({header, body}: {header: any, body: any}) => {
+                    let offence = false;
+
+                    // does msg have a body?
```
knocte, line 567: @realmarv I think this comment is wrong (it seems the next line after is not doing what the comment says)

commitlintplugins.test.ts
knocte: @realmarv let's add a new test that has the commit message "Revert .NET6 migration", and it should have assert(status).toBe(0)

commitlintplugins.test.ts
realmarv: Right. So should I find a way to set a higher value for body max length in that test? (Because now the limit is 64 and it will fail anyway) Also, what do you think about my suggestion:
> I say "This reverts commit bellow because bla bla" in the first line of the body and put the commit hash in the next line.

commitlint.config.ts

realmarv:
Should it be without a body?

knocte:
> Should it be without a body?

Yes, just a title that says "Revert .NET6 upd as it broke CI"

commitlintplugins.test.ts
knocte: I just want a commit message that includes the This reverts commit xyz. in the 2nd line, but with more text in the 2nd line. Because this is a testcase that the rule should not complain about, because the user has modified the default revert message.

realmarv:
> "Revert .NET6 upd as it broke CI"

Done

commitlint.config.ts
```diff
@@ -564,7 +564,7 @@ module.exports = {
                 'proper-revert-message': ({header, body}: {header: any, body: any}) => {
                     let offence = false;
 
-                    if (header.match(/^[Rr]evert/) !== null) {
+                    if (header.match(/^[Rr]evert ".+"/) !== null) {
```
knocte, line 567: @realmarv let's add another test with this title (no body): `Revert "process overhaul" to fix CI`

realmarv: It's a failing test, right? because I don't see the reason of reverting

commitlint.config.ts
```diff
@@ -333,7 +333,9 @@ module.exports = {
         'proper-issue-refs': [RuleStatus.Error, 'always'],
         'too-many-spaces': [RuleStatus.Error, 'always'],
         'commit-hash-alone': [RuleStatus.Error, 'always'],
+        'proper-revert-message': [RuleStatus.Error, 'always'],
     },
+    defaultIgnores: false,
```
knocte, line 338: @realmarv as it's not very clear why this is needed, let's add a comment that explains it; after that, I'll merge this PR

commitlint.config.ts
```diff
@@ -558,7 +560,28 @@ module.exports = {
...
+                'proper-revert-message': ({header, body}: {header: any, body: any}) => {
+                    let offence = false;
+
+                    if (header.match(/^[Rr]evert ".+"$/) !== null) {
+
+                        // does msg have a body?
+                        if (body !== null) {
+                            let bodyStr = convertAnyToString(body, "body").trim();
+                            let lines = bodyStr.split('\n');
+                            offence = lines[0].match(/^This reverts commit [^ ]{40}.$/) !== null && lines.length == 1;
```
knocte, line 573: @realmarv are you sure you don't need to escape the '.' here? AFAIU, dots in regexes act as any character, but here I think you're trying to match it with a real dot

commitlint.config.ts
realmarv: Done

commitlint.config.ts

knocte:
You didn't push your branch with push1by1. Also, this PR needs a rebase.

commitlintplugins.test.ts
knocte: is this resolved?

commitlint.config.ts
knocte: if done, click on "Resolve conversation" please

commitlintplugins.test.ts
realmarv: Yes

commitlint.config.ts

realmarv:
> You didn't push your branch with push1by1. Also, this PR needs a rebase.

Done

commitlint.config.ts
```diff
@@ -356,7 +356,11 @@ module.exports = {
         'too-many-spaces': [RuleStatus.Error, 'always'],
         'commit-hash-alone': [RuleStatus.Error, 'always'],
         'title-uppercase': [RuleStatus.Error, 'always'],
+        'proper-revert-message': [RuleStatus.Error, 'always'],
     },
+    // Commitlint automatically ignores some kinds of commits like Revert commit messages.
```
knocte, line 361: 2 last things to do in this PR:
- (Nit) Add an EOL before this comment above, to visually separate it better from the rule configuration list.
- Remove the TODO comment related to this PR, at line 363.

realmarv:
This PR is closed and replaced with https://github.com/nblockchain/conventions/pull/68 because some commits needed to be squashed and rebased.

knocte:
This PR should also remove one TODO line from commitling.config.ts.

knocte:
Oops, this one is closed, sorry.

PR #34, opened 2022-11-17
Ignore binary files and auto-generated folders in eofConventions.fsx

realmarv:


eofConvention.fsx
```diff
@@ -10,14 +10,22 @@ let EolAtEof(fileInfo: FileInfo) =
...
-    not (fileInfo.FullName.Contains $"%c{Path.DirectorySeparatorChar}%s{dirName}%c{Path.DirectorySeparatorChar}")
+let InDir (dirName: string) (fileInfo: FileInfo) = 
+    fileInfo.FullName.Contains $"%c{Path.DirectorySeparatorChar}%s{dirName}%c{Path.DirectorySeparatorChar}"
+
+let NotInDirs (dirNames: List<string>) (fileInfo: FileInfo) =
+    not (dirNames 
+        |> Seq.map (fun dirName -> InDir dirName fileInfo)
+        |> Seq.contains true)
+
+let whitelistExtensions = [".svg"; ".png"; ".slnf"; ".so"; ".a"; ".dll"; ".pdb"; ".dylib"]
```
knocte, line 21: @realmarv I said I want you to ignore all binary files, not create a binary file blacklist

realmarv: Done. 
Please note that in the "[FileConventions: implement HasBinaryContent](https://github.com/nblockchain/conventions/pull/34/commits/96b1d6a5e526de00163ee551d707442dd366270d)" commit the CI is red because eofConvention.fsx complains about binary files in DummyFiles (I used HasBinaryContent in the eofConvention.fsx later). if you click on the ci icon, the CI related to dotnet tests is green.

src/FileConventions/Library.fs
```diff
@@ -0,0 +1,10 @@
+﻿module FileConventions
+
+open System
+open System.IO
+open System.Linq
+
+let HasBinaryContent (fileInfo: FileInfo) =
+    use streamReader = new StreamReader (fileInfo.FullName)
+    let content = streamReader.ReadToEnd()
+    content.Any(fun ch -> Char.IsControl(ch) && ch <> '\r' && ch <> '\n')
```
knocte, line 10: @realmarv we already talked about how using .ReadToEnd() could have bad performance implications, let's use the same technique that you have learned before to prevent this, it's just 2 lines of code change (NOTE: I'm not expecting you now that, by default without asking me, you seek the algorithm with the best performance, because 99% of the time we don't need that; however if you have already learned how to prevent a performance pitfall that only takes a 2 LOC change and zero research, like in this case, please apply it without hesitation and without asking me.)

knocte: @realmarv nit: do not use the abbreviation "ch", there's no point in saving characters in source code

realmarv: > @realmarv we already talked about how using .ReadToEnd() could have bad performance implications, let's use the same technique that you have learned before to prevent this, it's just 2 lines of code change (NOTE: I'm not expecting you now that, by default without asking me, you seek the algorithm with the best performance, because 99% of the time we don't need that; however if you have already learned how to prevent a performance pitfall that only takes a 2 LOC change and zero research, like in this case, please apply it without hesitation and without asking me.)

In the previous case, we only needed the first line of the code. but in this case, I'm not sure how many lines are enough. So, should I iterate over the lines of the file until I find a control character?

knocte:
There are WIP commits in this PR which means the only way for me to merge it would be to squash all commits when merging, however in this case I shouldn't do that because you added things to gitignore, and that's not related to ignoring binary&autogenerated files in the eof script.

src/FileConventions/Library.fs
knocte: > I'm not sure how many lines are enough

You're using Any(), so **all** lines need to be iterated. But it doesn't matter, because ReadLines() goes line by line, instead of putting the whole file in memory (imagine if you try to read a 1GB this way).

realmarv:
> There are WIP commits in this PR which means the only way for me to merge it would be to squash all commits when merging, however in this case I shouldn't do that because you added things to gitignore, and that's not related to ignoring binary&autogenerated files in the eof script.

Should I just squash WIP commits? or should I squash all the commits related to ignoring binary files into one commit?

src/FileConventions/Library.fs

knocte:
In our company, squashing commits means just squashing WIP commits or commits that don't have any reason to exist on their own. It doesn't mean squashing commits into one (unless explicitly requested).

realmarv:
This PR is closed and replaced by [another PR](https://github.com/nblockchain/conventions/pull/35) because some commits needed to be squashed.

PR #35, opened 2022-11-23
Ignore binary files and auto-generated folders in eofConventions.fsx 

realmarv:


knocte:
There are many commits that don't have a CI run status, you need to push 1 by 1 with this script: https://github.com/nblockchain/fsx/blob/master/Tools/gitPush1by1.fsx

realmarv:
> There are many commits that don't have a CI run status, you need to push 1 by 1 with this script: https://github.com/nblockchain/fsx/blob/master/Tools/gitPush1by1.fsx

Done

knocte:
There's still one commit from the PR that doesn't have a CI status (the whitelist one).

Now let's talk about CI statuses, for example:
Screenshot 2022-11-23 at 9 53 35 PM

As you can see, the commit "create dotnet project" is actually failing for the same reason as the previous commit. It's better you move the "gitignore: add EOL at EOF" before it to achieve a green CI status in the "dotnet project". There might be more cases like this; if so, change the order.

realmarv:
> There's still one commit from the PR that doesn't have a CI status (the whitelist one).
> 
> Now let's talk about CI statuses, for example: Screenshot 2022-11-23 at 9 53 35 PM
> 
> As you can see, the commit "create dotnet project" is actually failing for the same reason as the previous commit. It's better you move the "gitignore: add EOL at EOF" before it to achieve a green CI status in the "dotnet project". There might be more cases like this; if so, change the order.

Done

knocte:
This PR needs to be rebased unfortunately.

realmarv:
> This PR needs to be rebased unfortunately.

Done

scripts/eofConvention.fsx
```diff
@@ -10,15 +12,21 @@ let EolAtEof(fileInfo: FileInfo) =
...
+let whitelistFolders = ["node_modules"; ".git"; "Debug"; "obj"; "bin"; "DummyFiles"]
 
 let invalidFiles = 
     Directory.GetFiles(".", "*.*", SearchOption.AllDirectories) 
     |> Seq.map (fun pathStr -> FileInfo pathStr)
-    |> Seq.filter (NotInDir "node_modules")
-    |> Seq.filter (NotInDir ".git")
-    |> Seq.filter (NotInDir "DummyFiles")
+    |> Seq.filter (NotInDirs whitelistFolders)
+    |> Seq.filter (fun fileInfo -> not (FileConventions.HasBinaryContent fileInfo))
```
knocte, line 29: @realmarv sorry, wrong order of commits; this commit that moves the logic from the .fsx file to FileConventions library has to be done before you write any unit test, before you fix the bug! otherwise the diff for the bugfix is not clear

realmarv: Shouldn't I first implement HasBinaryContent function and then call it in the script?

knocte: yes but we need the first commits (before creating any test) to move the logic from script to library; otherwise the moving of the code from one place to another is obfuscating the fix

realmarv:
This PR is closed and superseded by https://github.com/nblockchain/conventions/pull/74

scripts/eofConvention.fsx
realmarv: Done

PR #37, opened 2022-11-25
A new rule that detects starting with lowercase letter in the titles without area

realmarv:


commitlintplugins.test.ts
```diff
@@ -423,6 +423,27 @@ test('subject-lowercase10', () => {
...
+    let titleUpperCase2 = runCommitLintOnMsg(commitMsgWithoutArea);
+    expect(titleUpperCase2.status).toBe(0);
+});
+
+
+test('title-uppercase3', () => {
+    let commitMsgWithoutArea = "testFixtureSetup refactor";
+    let titleUpperCase3 = runCommitLintOnMsg(commitMsgWithoutArea);
+    expect(titleUpperCase3.status).toBe(0);
+});
```
knocte, line 444: let's add another test: title-uppercase4 which has this title "lowercase: area is lowercase" whose status should be 0.

realmarv: Done

realmarv: Done

PR #39, opened 2022-12-26
Fix truncated body issue 

realmarv:


realmarv:
This PR is closed and replaced with https://github.com/nblockchain/conventions/pull/40 because some commits needed to be squashed.

PR #40, opened 2022-12-26
Fix truncated body issue

realmarv:


knocte:
Why does the test not have an empty 2nd line? we don't want a test to fail for the wrong reason.

realmarv:
> Why does the test not have an empty 2nd line? we don't want a test to fail for the wrong reason.

Sorry fixed it

knocte:
Why are you using backticks for the message in the test instead of double-quotes like other tests?

knocte:
Please squash the last commit in the proper place.

knocte:
> Sorry fixed it

@realmarv you fixed it in this PR but not in the commitlint bug.

realmarv:
> > Sorry fixed it
> 
> @realmarv you fixed it in this PR but not in the commitlint bug.

Done

PR #41, opened 2022-12-27
Fix body-prose empty paragraph bug

realmarv:


knocte:
Why do you use backticks again for the message? Let's use double-quotes (like other tests), and this way you don't need to escape the backticks inside.

realmarv:
> Why do you use backticks again for the message? Let's use double-quotes (like other tests), and this way you don't need to escape the backticks inside.

Ah sorry, the initial commit message is too long, I forgot to replace the backtick with double quotes at the end.

PR #43, opened 2022-12-28
Fix footer-notes-misplacement truncated body

realmarv:


knocte:
Good work! BTW have you seen this? https://github.com/nblockchain/conventions/actions/runs/3787410150/jobs/6439186147 I think we need to ignore title length limit for commits that start with "Merge ", what do you think?

realmarv:
> Good work! BTW have you seen this? https://github.com/nblockchain/conventions/actions/runs/3787410150/jobs/6439186147 I think we need to ignore title length limit for commits that start with "Merge ", what do you think?

Right. I'll take care of it.

PR #44, opened 2022-12-28
Ignore merge commits in header max length rule

realmarv:


commitlint.config.ts
```diff
@@ -437,7 +437,7 @@ module.exports = {
                     let offence = false;
 
                     let message = `Please do not exceed ${maxLineLength} characters in title.`;
-                    if (headerStr.length > maxLineLength) {
+                    if (!headerStr.startsWith('Merge') && headerStr.length > maxLineLength) {
```
knocte, line 440: should be 'Merge '

PR #54, opened 2023-01-30
Upgrade commitlint

realmarv:

PR #51, opened 2023-01-12
Upgrade commitlint to 17.4.0 version without upgrading typescript to 4.9.4

realmarv:

PR #55, opened 2023-01-30
Upgrade commitlint to the latest version

realmarv:
Replaced https://github.com/nblockchain/conventions/pull/54.


.github/workflows/ci.yml
```diff
@@ -58,6 +58,10 @@ jobs:
           git --version
           node --version
           npm --version
+      - name: Install commitlint
+        run: npm install
+      - name: Print commitlint version
+        run: npx commitlint --version
```
knocte, line 64: @realmarv this change seems redundant because commitlint.sh (which is called in the next step) already calls these commands above

knocte:
Please expand the commit msg of the 2nd commit: you want to add as much info as possible to explain the **why**. In this case, you may want to say, for example, that the workaround is not needed anymore because we're now using a recent-enough version of commitlint that includes the fix for the bug, and put the URL to the bug, etc

realmarv:
> Please expand the commit msg of the 2nd commit: you want to add as much info as possible to explain the **why**. In this case, you may want to say, for example, that the workaround is not needed anymore because we're now using a recent-enough version of commitlint that includes the fix for the bug, and put the URL to the bug, etc

Done

knocte:
This PR is still not upgrading the TypeScript compiler version, so this PR should be named in the same way as #51 (and the 1st commit title should include this info too).

However, I've just discovered something else: if you upgrade TypeScript, we don't get those weird compiler errors anymore (which possibly means that they were a typescript compiler bug that has been recently fixed).

That being said, upgrading TS compiler is still not succeeding, but it's because of a compiler error that seems to be easy to fix. So let's upgrade everything properly in this PR (do it in a 3rd commit first, and after you see that it's green, you can probably squash it with the 1st).

knocte:
> and after you see that it's green, you can probably squash it with the 1st).

Actually no need to squash, there may be value in keeping that first commit intact, but it should be renamed to say that TS version is not being upgraded.

.github/workflows/ci.yml
```diff
@@ -58,6 +58,19 @@ jobs:
           git --version
           node --version
           npm --version
+          npx typescript --version
+      - name: Upgrade typescript
+        run: npm install -g typescript@4.9.5
```
knocte, line 63: @realmarv this is not exactly what I mean when I recommended upgrading typescript; I didn't say "upgrade typescript explicitly", I meant to have latest typescript in the same way we had it before I committed the downgrade to commitlint17.0.0: by not having any package*.json files at all.

knocte:
So the title of this PR is outdated I guess? (After you landed your last 3rd commit.)

knocte:
BTW the commit msg of 3rd commit is wrong: you're saying that commit-lint is being upgraded to 17.4.2 but that was already the case, that was already done by the 1st commit.

realmarv:
> So the title of this PR is outdated I guess? (After you landed your last 3rd commit.)

Done

knocte:
After I merged the Prettier PR, there are conflicts here, sorry. Can you rebase it? Thanks

realmarv:
> After I merged the Prettier PR, there are conflicts here, sorry. Can you rebase it? Thanks

Done

knocte:
And this needs another rebase :sob:

realmarv:
> And this needs another rebase sob

Done

.github/workflows/CI.yml
```diff
@@ -25,7 +28,12 @@ jobs:
           npx commitlint --version
       - name: Install yarn
         run: |
-          sudo npm install -g yarn
+          npm install -g yarn
+          yarn add --dev jest typescript ts-jest @types/jest
+      - name: Install commitlint
+        run: |
+          npm install conventional-changelog-conventionalcommits
+          npm install commitlint@17.4.2
```
knocte, line 36: @realmarv isn't this already done in commitlint.sh? so we don't need it here

.github/workflows/CI.yml
```diff
@@ -58,6 +71,7 @@ jobs:
           git --version
           node --version
           npm --version
+          npx commitlint --version
```
knocte, line 74: @realmarv isn't this already done by commitlint.sh?

.github/workflows/CI.yml
```diff
@@ -14,9 +14,12 @@ jobs:
           apt update
           apt install -y sudo
           sudo apt install -y git
-          sudo DEBIAN_FRONTEND=noninteractive apt install --yes npm
-      - name: Install commitlint
-        run: npm install
+
+          sudo apt install -y curl
+          # can't install ubuntu's default nodejs version because we would get this error:
+          # error @commitlint/config-conventional@17.1.0: The engine "node" is incompatible with this module. Expected version ">=v14". Got "12.22.9"
```
knocte, line 20: @realmarv this is not the same error explained in the commit msg

realmarv: It seems there is no error in the sanitycheck job

.github/workflows/CI.yml
```diff
@@ -44,7 +52,12 @@ jobs:
         run: |
           apt update && apt install -y sudo
           sudo apt install -y git
-          sudo DEBIAN_FRONTEND=noninteractive apt install --yes npm
+
+          sudo apt install -y curl
+          # can't install ubuntu's default nodejs version because we would get this error:
+          # error @commitlint/config-conventional@17.1.0: The engine "node" is incompatible with this module. Expected version ">=v14". Got "12.22.9"
```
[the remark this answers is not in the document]
realmarv: It seems there is no error in the sanitycheck job

.github/workflows/CI.yml
realmarv: Right fixed it

knocte:
@realmarv please click on "resolve conversation" for things you have already addressed.

.github/workflows/CI.yml

.github/workflows/CI.yml
realmarv: We don't run commitlint.sh in tests job

commitlint/helpers.ts
```diff
@@ -22,13 +22,7 @@ export abstract class Helpers {
...
         if (potentialString === null || potentialString === undefined) {
             // otherwise, String(null) might give us the stupid string "null"
-            throw new Error(
-                "Unexpected " +
-                    paramName +
-                    "===null or " +
-                    paramName +
-                    "===undefined happened"
-            );
+            return "";
```
knocte, line 25: @realmarv this needs to be fixed; already discussed in other PR

commitlint/helpers.ts
```diff
@@ -19,16 +19,10 @@ export abstract class Helpers {
     public static convertAnyToString(
         potentialString: any,
         paramName: string
-    ): string {
+    ): any {
```
knocte, line 22: @realmarv look at the name of this function, the point of it is to not deal with an `any` object! null should be a valid string, and if it's not, then you can use `string?` I think, not sure if it's `?` or other char

realmarv: Done

commitlint/helpers.ts
realmarv: Done

commitlint/plugins.ts
```diff
@@ -369,18 +431,13 @@ export abstract class Plugins {
     }
 
     public static bodySoftMaxLineLength(
-        rawStr: string,
+        bodyStr: string | null,
```
knocte, line 434: @realmarv I understand that in this commit you need to change the rules that receive bodyStr because there might be no body (so, null); however, the other rules that don't deal with the body should not change, in those cases you can assert that rawStr is not null and they receive string, not `string | null`.

realmarv: The output of the convertAnyToString function can be null or string and its output is the input to the plugins. What do you suggest me to do?

knocte: in the cases where you call convertAnyToString with raw, cast the result from `string|null` to a `string` (not sure if this is done via an assert)

realmarv: I'm not sure an assert works but I can use casting like this `(Helpers.convertAnyToString(raw, "raw") as string)`. What do you think?

knocte: This is still not resolved. By the way, I checked the log of the last CI run and it is using 17.4.3 so you should rename this PR to not specify 17.4.3, maybe just say "latest".

knocte: typo in above msg ^ , edited it

commitlint/helpers.ts
```diff
@@ -19,18 +19,28 @@ export abstract class Helpers {
...
-                    paramName +
-                    "===null or " +
-                    paramName +
-                    "===undefined happened"
+                "This is unexpected because raw should never be null"
+            );
+        }
+    }
+
+    public static assertHeader(headerStr: string | null) {
```
knocte, line 38: @realmarv good work! this PR is almost ready, but I just realised that, if you're using this assertFoo functions, you can improve 2 things here:
- Have a generic single function instead of 2 functions, call it assertNotNull, pass the error message as 2nd argument.
- Have this function return string, and do the cast with `as` inside the function.

.github/workflows/CI.yml
knocte, line 36: @realmarv this is the last thing to fix, it shouldn't specify 17.4.2 because it doesn't work (CI says it's in 17.4.3 anyway). Fix also the commit msg when fixing this.

commitlint/helpers.ts
realmarv: Done

.github/workflows/CI.yml
realmarv: Fixed the version. But I guess the version in the third commit message is correct because I looked at the CI of that commit and the commitlint version is 17.4.2.

knocte: > I looked at the CI of that commit and the commitlint version is 17.4.2.

what? 🤔 I don't understand that, are you sure you're referring to the last commit?

knocte: I just double-checked; one job uses 17.4.2 and the other 17.4.3. This is not acceptable, either both use 17.4.2 or both use 17.4.3.

realmarv: Not the last commit. The third commit. I fixed the commit message of the last commit

knocte: I'm even more confused now, because there are 3 commits in this PR; therefore the last commit is the third commit.

realmarv: Nevermind I guess it's done right?

knocte: Yes, but being picky, the 3rd commit's title should not just mention "dependencies", because it's upgrading commitlint too.

knocte: Oh, you fixed that already! nice

knocte:
@realmarv you forgot to remove the workaround in the bodyProse() plugin, please create a new PR.

realmarv:
> @realmarv you forgot to remove the workaround in the bodyProse() plugin, please create a new PR.

Because bodyProse workaround is for a different issue

knocte:
Ah you're right sorry.

knocte:
BTW please prepare a commitlint branch that is updated to latest master and contains your commit that adds GitHubCI please (I will review before you post PR).

Issue #46, opened 2023-01-03
Suggestions shouldn't be offered for the area


knocte:
Testcase:

```
configure: this is the very very very very very long title
```

Error message should NOT recommend the "config" suggestion.

PR #56, opened 2023-01-30
No replacement suggestions for the area

realmarv:
Fixes https://github.com/nblockchain/conventions/issues/46

### DIFF
--- a/commitlint.config.ts
+++ b/commitlint.config.ts
@@ -308,11 +308,19 @@ module.exports = {
                     let message = `Please do not exceed ${maxLineLength} characters in title (found ${headerLength}).`;
                     if (!headerStr.startsWith('Merge ') && headerLength > maxLineLength) {
                         offence = true;
+
+                        let colonIndex = headerStr.indexOf(':');
+
+                        let titleWithoutArea = headerStr;
+                        if (colonIndex > 0){
+                            titleWithoutArea = headerStr.substring(colonIndex);
+                        }
+                        
                         let numRecomendations = 0;
-                        let lowerCaseHeaderStr = headerStr.toLowerCase()
+                        let lowerCaseTitleWithoutArea = titleWithoutArea.toLowerCase();
                         Object.entries(abbr).forEach(([key, value]) => {  
                             let pattern = new RegExp("\\b(" + key.toString() + ")\\b")
-                            if (pattern.test(lowerCaseHeaderStr)){
+                            if (pattern.test(lowerCaseTitleWithoutArea)){
                                 if (numRecomendations === 0) {
                                     message = message + ' The following replacement(s) in your commit title are recommended:\n'
                                 }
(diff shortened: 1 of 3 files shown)


knocte:
I would prefer if you added, as a first commit, the configure -> config suggestion. And after that, in the second commit where you introduce the failing test, use "configure:" instead of "configuration:" in the test (reason for this is that usually "configure" is the name of a script). BTW let's rebase this PR too.

commitlint.config.ts
```diff
@@ -300,10 +300,18 @@ module.exports = {
...
+                        let colonIndex = headerStr.indexOf(':')
+
+                        let title = headerStr
+                        if (colonIndex > 0){
+                            title = headerStr.substring(colonIndex)
+                        }
+                        
                         let numRecomendations = 0;
-                        let lowerCaseHeaderStr = headerStr.toLowerCase()
+                        let lowerCaseTitle = title.toLowerCase()
```
knocte, line 312: @realmarv missing semicolon in many places above ^

commitlint.config.ts
```diff
@@ -300,10 +300,18 @@ module.exports = {
                     let message = `Please do not exceed ${maxLineLength} characters in title (found ${headerLength}).`;
                     if (!headerStr.startsWith('Merge ') && headerLength > maxLineLength) {
                         offence = true;
+
+                        let colonIndex = headerStr.indexOf(':')
+
+                        let title = headerStr
+                        if (colonIndex > 0){
+                            title = headerStr.substring(colonIndex)
```
knocte, line 308: @realmarv a better name would be `titleWithoutArea`

commitlint.config.ts

realmarv:
> I would prefer if you added, as a first commit, the configure -> config suggestion. And after that, in the second commit where you introduce the failing test, use "configure:" instead of "configuration:" in the test (reason for this is that usually "configure" is the name of a script). BTW let's rebase this PR too.

Done

commitlint.config.ts
```diff
@@ -300,12 +300,20 @@ module.exports = {
...
+                        
                         let numRecomendations = 0;
-                        let lowerCaseHeaderStr = headerStr.toLowerCase()
+                        let lowerCaseTitleWithoutArea = titleWithoutArea.toLowerCase();
                         Object.entries(abbr).forEach(([key, value]) => {  
-                            if (lowerCaseHeaderStr.includes(key.toString())){
+                            if (lowerCaseTitleWithoutArea.includes(key.toString())){
                                 if (numRecomendations === 0) {
-                                    message = message + ' The following replacement(s) in your commit title are recommended:\n'
+                                    message = message + ' The following replacement(s) in your commit title are recommended:\n';
```
knocte, line 316: @realmarv this line doesn't need to change in this PR (I know it's missing a semicolon, but it works; and if I let you change it, then I would be ok by making you arbitrary style changes in any other part of the file(s), which is not acceptable for a non-refactoring PR); sorry to be picky, but the point of this is to prevent git-blame noise; and by using Prettier soon (thanks to your other PR), we will avoid these problems later by fixing the root cause

realmarv: Right. Done

knocte:
Can you rebase this please? While you're at it, fix one of the commit messages (abbreviations is not an area). Cheers

Issue #49, opened 2023-01-10
Adopt "prettier" for our TypeScript code


knocte:
In the same way we use fantomless/fantomas for F# code, we should use Prettier tool for TypeScript.

FTR: https://www.npmjs.com/package/prettier

PS: Let's not use https://github.com/marketplace/actions/prettier-action , we don't want to depend on GitHubActions marketplace.

PR #59, opened 2023-01-30
Adopt "prettier" for our TypeScript code

realmarv:
Fixes https://github.com/nblockchain/conventions/issues/49

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,4 +1,4 @@
-[*.{fs,fsx}]
+[*.{fs,fsx,ts}]
 end_of_line=lf
 indent_size=4
 max_line_length=80
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -68,3 +68,20 @@ jobs:
         run: sudo apt install -y dotnet6
       - name: Run script to make sure all files end with EOL 
         run: dotnet fsi eofConvention.fsx
+      - name: Install prettier
+        run: npm install prettier@2.8.3
+      - name: Change file permissions
+        # We need this step so we can change the files using `npx prettier --write` in the next step.
+        # Otherwise we get permission denied error in the CI.
+        run: sudo chmod 777 -R .
+      - name: Run "prettier" to check the style of our TypeScript code
+        run: |
+          sudo npx prettier --quote-props=consistent --write './**/*.ts'
+          # Since we changed file modes in the previous step we need the following command to
+          # make git ignore mode changes in files and doesn't include them in the git diff command.
+          git config core.fileMode false
+          # Since after installing npm packages, package.json and package-lock.json change, 
+          # `git diff --exit-code` will always throw an error, so we need to restore those
+          # files before running `git diff --exit-code` command.
+          git restore package.json package-lock.json
+          git diff --exit-code
--- a/commitlint/abbreviations.ts
+++ b/commitlint/abbreviations.ts
@@ -252,7 +252,7 @@ export let abbr = {
     "timers": "tmrs",
     "transaction": "tx",
     "transactions": "txs",
-    "two": 2,
+    "two": "2",
     "unix to dos": "unix2dos",
     "unnecessary": "unneeded",
     "unsigned integer": "uint",
@@ -266,4 +266,4 @@ export let abbr = {
     "yes/no": "y/n",
     "zero": "0",
     "zeroes": "0s",
-}
+};
--- a/commitlint.config.ts
+++ b/commitlint.config.ts
@@ -4,7 +4,13 @@ import { abbr } from "./commitlint/abbreviations";
 function convertAnyToString(potentialString: any, paramName: string): string {
     if (potentialString === null || potentialString === undefined) {
         // otherwise, String(null) might give us the stupid string "null"
-        throw new Error('Unexpected ' + paramName + '===null or ' + paramName + '===undefined happened');
+        throw new Error(
+            "Unexpected " +
+                paramName +
+                "===null or " +
+                paramName +
+                "===undefined happened"
+        );
     }
     return String(potentialString);
 }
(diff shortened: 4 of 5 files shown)


knocte:
Let's not deviate from our current style, please configure prettier with 4-space indentation, not 2-space.

.github/workflows/ci.yml
```diff
@@ -68,3 +68,5 @@ jobs:
         run: sudo apt install -y dotnet6
       - name: Run script to make sure all files end with EOL 
         run: dotnet fsi eofConvention.fsx
+      - name: prettier
+        run: npx prettier --check --tab-width=4 './**/*.ts'
```
knocte, line 72: isn't there a way to specify tab width via .editorconfig file?

knocte: BTW this is very brittle because it will always use the last prettier version; we should pin to the current last version now to avoid CI break on us everytime there's a new version that changes the style

.github/workflows/ci.yml
```diff
@@ -68,3 +68,5 @@ jobs:
         run: sudo apt install -y dotnet6
       - name: Run script to make sure all files end with EOL 
         run: dotnet fsi eofConvention.fsx
+      - name: prettier
```
knocte, line 71: @realmarv please use a better name, see other name of steps

commitlint/abbreviations.ts
```diff
@@ -1,186 +1,186 @@
...
-    "buffer": "buf",
-    "button": "btn",
-    "calculate": "calc",
-    "callback": "cb",
-    "channel": "chan",
-    "character": "char",
-    "collection": "coll",
-    "column": "col",
-    "command": "cmd",
+    absolute: "abs",
```
knocte, line 4: @realmarv woah, seriously? I can't believe prettier would do this by default, remove the quotes for single words and leave the quotes for words with spaces? my god, can you investigate how to disable this behaviour please, we want all of them with double quotes

realmarv: Done

.github/workflows/ci.yml
knocte: `Adopt "prettier" for our TypeScript code` is a good name for a commit or PR title, but for a CI step? Once this PR gets merged, then we have adopted prettier, but it only happens once. The step "runs prettier", it doesn't adopt it.

commitlint/abbreviations.ts
knocte: Nice.

.github/workflows/ci.yml

.editorconfig
```diff
@@ -1,4 +1,4 @@
-[*.{fs,fsx}]
+[*.{fs,fsx,ts}]
```
knocte, line 1: this is wrong because most settings that are here don't apply to TS files but to FS

knocte: oh, but those are prefixed with fsharp_, mmm, nevermind then, this shoudl work

.github/workflows/ci.yml
```diff
@@ -68,3 +68,7 @@ jobs:
         run: sudo apt install -y dotnet6
       - name: Run script to make sure all files end with EOL 
         run: dotnet fsi eofConvention.fsx
+      - name: Install prettier
+        run: npm install prettier@2.8.3
+      - name: Run "prettier" to style our TypeScript code
```
knocte, line 73: @realmarv yes this step runs prettier, but it's not to style our TypeScript code, it is to check if the person that has committed his work has run prettier before committing

.github/workflows/ci.yml
```diff
@@ -68,3 +68,7 @@ jobs:
...
+        run: npx prettier --quote-props=consistent --check './**/*.ts'
```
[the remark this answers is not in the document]

.github/workflows/ci.yml

.github/workflows/ci.yml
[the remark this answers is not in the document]
realmarv: I guess it's done now

.github/workflows/CI.yml
```diff
@@ -68,3 +68,14 @@ jobs:
...
+      - name: Install prettier
+        run: npm install prettier@2.8.3
+      - name: Change file permissions
+        # We need this step so we can change the files using `npx prettier --write` in the next step.
+        # Otherwise we get permission denied error in the CI.
+        run: sudo chmod 777 -R .
+      - name: Run "prettier" to check the style of our TypeScript code
+        run: |
+          sudo npx prettier --quote-props=consistent --write './**/*.ts'
+          git config core.fileMode false
```
knocte, line 80: @realmarv I suggested this workaround so that you dont need to do weird things with chmod; either use this or chmod but not both

.github/workflows/CI.yml
```diff
@@ -68,3 +68,14 @@ jobs:
...
+          git diff --exit-code './**/*.ts'
```
knocte, line 81: @realmarv let's revisit this conversation sorry; if we increase the use of prettier then I'm not sure git diff can receive many arguments; so, just in case (and because it's more readable and self-explanatory anyway), let's rather use `git restore` (with the package.json files as arg) before `git diff` command.

.github/workflows/CI.yml
realmarv: I thought this command makes git ignore mode changes in files and doesn't include it in git diff command. How does it solve permission denied error?

knocte:
> ci: check TS files coding style

FYI I renamed "ci.yml" to "CI.yml" so, nit: let's fix the area name. It is an acronym so it has to be uppercase.

.github/workflows/CI.yml
knocte: oh, I thought the chmod thing was solving git diff showing mask file differences, my bad; anyway, then again you need to explain with a comment what that line is doing because it's not self-explanatory

.github/workflows/CI.yml
realmarv: Doesn't `git restore` command reset all files to the last commit state?

knocte: Read my comment again, I said **git restore (with the package.json files as arg)**

knocte:
> Prettify TS files coding style using prettier tool

This commit should have an area: `commitlint`.

realmarv:
> > Prettify TS files coding style using prettier tool
> 
> This commit should have an area: `commitlint`.

Done

.github/workflows/CI.yml

Issue #60, opened 2023-01-31
Suggestions should not be offered for sub-words


knocte:
Testcase:

```
Fix android build because blah blah very very very very very long title
```

Error message should NOT recommend the "and" -> "&" suggestion.

PR #61, opened 2023-01-31
Stop offering replacement suggestions for sub-words

realmarv:
Fixes https://github.com/nblockchain/conventions/issues/60


knocte:
As I said in PR#57, let's improve the commit messages, e.g. include the "Fixes ..." sentence in the second commit, things like that.

commitlint.config.ts
```diff
@@ -303,7 +303,8 @@ module.exports = {
                         let numRecomendations = 0;
                         let lowerCaseHeaderStr = headerStr.toLowerCase()
                         Object.entries(abbr).forEach(([key, value]) => {  
-                            if (lowerCaseHeaderStr.includes(key.toString())){
+                            let pattern = new RegExp("\\b(" + key.toString() + ")\\b")
```
knocte, line 306: @realmarv are we sure this doesn't break other scenarios? to make sure, let's add these 2 tests in a 3rd commit:

a) ```title: 1 second``` should offer `1sec` suggestion
b) ```Configuration simplification``` should offer `config`

realmarv:
> As I said in PR#57, let's improve the commit messages, e.g. include the "Fixes ..." sentence in the second commit, things like that.

Done

commitlintplugins.test.ts
```diff
@@ -448,10 +448,10 @@ test('header-max-length-with-suggestions5', () => {
 test('header-max-length-with-suggestions6', () => {
     let commitMsgWithThatExceedsHeaderMaxLength =
         "Upgrade foo bla bla bla bla bla bla bla bla bla bla bla bla bla bla";
-    let headerMaxLength1 = runCommitLintOnMsg(commitMsgWithThatExceedsHeaderMaxLength);
+    let headerMaxLength6 = runCommitLintOnMsg(commitMsgWithThatExceedsHeaderMaxLength);
     let expected_message = `"upgrade" -> "update"`
-    expect(headerMaxLength1.status).not.toBe(0);
-    expect((headerMaxLength1.stdout + '').includes(expected_message)).toEqual(true)
+    expect(headerMaxLength6.status).not.toBe(0);
+    expect((headerMaxLength6.stdout + '').includes(expected_message)).toEqual(true)
```
knocte, line 454: @realmarv the commit msg says "add new tests", but you're modifying tests now; sorry to ask this question again, are you using `git gui` to commit? if you were using git gui, you would find this before you commit

realmarv: I use gitk. I guess it's more related to not changing other parts of the code in non-refactoring PR.

knocte: > it's more related to not changing other parts of the code in non-refactoring PR.

It's BOTH. You should not change other parts of the code in non-refactoring commit, but to catch this mistake early **you need a visual tool to commit your changes**, and that's why I recommend using git gui, to see the diff before you commit it. If gitk doesn't allow you to see the diff at commit time, stop using it now for that.

realmarv: Ok no problem. I was using vscode git to see the diff in commit time

knocte:
Let's add a 4th commit that has two more tests:

a) "area: 20 characters more" -> should offer "characters" -> "chars"
b) "Split that compares better" -> should **not** offer "compares" -> "cmps"

realmarv:
> Let's add a 4th commit that has two more tests:
> 
> a) "area: 20 characters more" -> should offer "characters" -> "chars" b) "Split that compares better" -> should **not** offer "compares" -> "cmps"

characters is not in appreviations

commitlint.config.ts
realmarv: Done

knocte:
> characters is not in appreciation. Do I add it?

Well, that would fix only this specific test, but what I wanted to highlight is that this PR, if merged as is, would break all plurals. So we should fix all plurals (not just this one), before merging.

realmarv:
> > characters is not in appreciation. Do I add it?
> 
> Well, that would fix only this specific test, but what I wanted to highlight is that this PR, if merged as is, would break all plurals. So we should fix all plurals (not just this one), before merging.

Fixing it how? by adding all plurals to the list?

knocte:
> by adding all plurals to the list?

Is there any other way?

realmarv:
> > by adding all plurals to the list?
> 
> Is there any other way?

For example, marking each of the verbs and nouns as being verb or noun and considering nouns with or without s. something like this. But I think just adding plurals is easier

knocte:
> But I think just adding plurals is easier

Exactly.

realmarv:
> > But I think just adding plurals is easier
> 
> Exactly.

Done

knocte:
> abbreviations: add plural form of nouns

`abbreviations` is not an area, it's a sub-area; the area is commitlint

realmarv:
> > abbreviations: add plural form of nouns
> 
> `abbreviations` is not an area, it's a sub-area; the area is commitlint

Done

knocte:
> Done

You've fixed 1 commit, there were 2 commits with same nit.

PR #64, opened 2023-02-03
Split helper functions and plugins from commitlint config

realmarv:


knocte:
> Split helper functions and plugins from commitlint config

Come on Zahra, at this point I would expect you to not forget an area and not forget that we have a 50char limit for the title.

realmarv:
> > Split helper functions and plugins from commitlint config
> 
> Come on Zahra, at this point I would expect you to not forget an area and not forget that we have a 50char limit for the title.

Sorry fixed it

knocte:
After I merged the Prettier PR, there are conflicts here, sorry. Can you rebase it? Thanks

PR #67, opened 2023-02-08
 A rule that rejects obvious words in the commit title

tehraninasab:
Replaced https://github.com/nblockchain/conventions/pull/15

### DIFF
--- a/commitlint/plugins.test.ts
+++ b/commitlint/plugins.test.ts
@@ -846,6 +846,68 @@ test("proper-issue-refs5", () => {
     expect(properIssueRefs5.status).not.toBe(0);
 });
 
+test("reject-obvious-words1", () => {
+    let commitMsgWithObviousWordAfterColon = "foo: change";
+    let rejectObviousWords1 = runCommitLintOnMsg(
+        commitMsgWithObviousWordAfterColon
+    );
+    expect(rejectObviousWords1.status).not.toBe(0);
+});
+
+test("reject-obvious-words2", () => {
+    let commitMsgWithObviousWordAfterColon = "foo: change bla bla";
+    let rejectObviousWords2 = runCommitLintOnMsg(
+        commitMsgWithObviousWordAfterColon
+    );
+    expect(rejectObviousWords2.status).not.toBe(0);
+});
+
+test("reject-obvious-words3", () => {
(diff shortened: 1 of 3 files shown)


knocte:
Let's consider the commitlint PR for fixing subject-full-stop highest priority, but after you've tackled that, you can address the feedback that I'm going to share about this PR.

commitlint/plugins.test.ts
```diff
@@ -652,6 +652,30 @@ test("proper-issue-refs3", () => {
...
+    expect(rejectObviousWord2.status).not.toBe(0);
+});
+
+test("reject-obvious-words3", () => {
+    let commitMsgWithoutObviousWordAfterColon = "foo: bla bla bla";
+    let rejectObviousWord3 = runCommitLintOnMsg(
+        commitMsgWithoutObviousWordAfterColon
+    );
+    expect(rejectObviousWord3.status).toBe(0);
+});
```
knocte, line 677: @realmarv I've slightly changed my mind of how this rule should work, but instead of explaining it, I'll just request you to add the following tests:
- `foo: update bla bla` -> status should not be 0
- `foo: update` -> status should be 0
- `foo: change` -> status should not be 0
- `foo: update bla bla\n\nBla blah.` -> status should be 0
- `foo: change bla bla\n\nBla blah.` -> status should not be 0

That is, there is an exception to the rule: the word update can go alone after area ("change" or "modify" cannot). The word `update` can only go with extra words after it if body length is higher than zero ("change" or "modify" cannot).

The above means that you should not make this rule extensible (i.e. by putting the list of words in a separate file or list). You have to hardcode them in the plugin instead.

realmarv: Done

knocte:
Please add a new failing test with this commit message: https://github.com/realmarv/conventions/commit/006fe222f6911bb3a522325388f6ee339e986120

That is, if the subject contains just the word "fix", and there's no commit msg body, this rule should complain as well.

realmarv:
> Please add a new failing test with this commit message: [realmarv@006fe22](https://github.com/realmarv/conventions/commit/006fe222f6911bb3a522325388f6ee339e986120)
> 
> That is, if the subject contains just the word "fix", and there's no commit msg body, this rule should complain as well.

Done

PR #68, opened 2023-02-08

realmarv:
Replaced https://github.com/nblockchain/conventions/pull/32


commitlint/helpers.ts
```diff
@@ -22,13 +22,7 @@ export abstract class Helpers {
...
         if (potentialString === null || potentialString === undefined) {
             // otherwise, String(null) might give us the stupid string "null"
-            throw new Error(
-                "Unexpected " +
-                    paramName +
-                    "===null or " +
-                    paramName +
-                    "===undefined happened"
-            );
+            return "";
```
knocte, line 25: @realmarv this is unacceptable, it should return null, not `""`

commitlint/plugins.ts
```diff
@@ -272,6 +272,24 @@ export abstract class Plugins {
...
 
+    public static properRevertMessage(headerStr: string, bodyStr: string) {
+        let offence = false;
+
+        if (headerStr.match(/^[Rr]evert ".+"$/) !== null) {
+            // does msg have a body?
+            if (bodyStr !== "") {
+                let lines = bodyStr.split("\n");
+                offence =
+                    lines[0].match(/^This reverts commit [^ ]{40}\.$/) !==
```
knocte, line 283: @realmarv 40? sorry, what is this magic number?

commitlint/plugins.ts
```diff
@@ -272,6 +272,24 @@ export abstract class Plugins {
...
+                        null && lines.length == 1;
```
knocte, line 284: `lines.length == 1` should go before accessing `lines[0]` or this might throw an exception

commitlint/plugins.ts
```diff
@@ -272,6 +272,24 @@ export abstract class Plugins {
...
+            } else {
+                offence = true;
+            }
+        }
+
+        return [!offence, `Please explain why you're reverting`];
```
knocte, line 290: @realmarv I've changed recently all error messages of all plugins, let's now be consistent with them here ^

commitlint.config.ts
```diff
@@ -43,7 +43,12 @@ module.exports = {
         "too-many-spaces": [RuleConfigSeverity.Error, "always"],
         "commit-hash-alone": [RuleConfigSeverity.Error, "always"],
         "title-uppercase": [RuleConfigSeverity.Error, "always"],
+        "proper-revert-message": [RuleConfigSeverity.Error, "always"],
```
knocte, line 46: @realmarv let's rename this rule to `default-revert-message` and change from `always` to `never`; then we should map the always/never argument properly to make `offence` be depending on it

commitlint/helpers.ts
realmarv: If I return null here I have to add extra indentation in all rules for `if (xStr !== null)` condition. Is that ok?

commitlint/plugins.ts
realmarv: The length of git commit hash

commitlint/helpers.ts
knocte: that is ok; what is not ok is masquerading a missing string with an empty string; they are not the same thing

commitlint/plugins.ts
knocte: it's not self-explanatory so you know what to do

commitlint/plugins.ts
realmarv: Since I it's in `if (bodyStr !== "")` block there is at least one line. The purpose of `lines.length == 1` condition is to check if there isn't more than one line because if there is, the revert message has definitely changed.

knocte: still, placing the length check before the match is valuable because of 1. readability and 2. performance (micro-optimization I know)

knocte:
This PR should also remove one TODO line from commitling.config.ts.

commitlint/plugins.ts
```diff
@@ -272,6 +272,28 @@ export abstract class Plugins {
...
 
+    public static defaultRevertMessage(headerStr: string, bodyStr: string) {
+        let offence = false;
+
+        if (headerStr.match(/^[Rr]evert ".+"$/) !== null) {
+            // does msg have a body?
+            if (bodyStr !== null) {
+                let lines = bodyStr.split("\n");
+                offence =
+                    // 40 is the length of git commit hash in the following regex pattern.
```
knocte, line 283: @realmarv nit, if you place the comment one line down more (in L284 instead of L283) then you don't need to explain "in the following regex"

commitlint/helpers.ts
```diff
@@ -19,16 +19,10 @@ export abstract class Helpers {
     public static convertAnyToString(
         potentialString: any,
         paramName: string
-    ): string {
+    ): any {
```
knocte, line 22: @realmarv I didn't request this change in this PR

realmarv: One of the default-revert-message rule inputs is body, so I needed bodyStr in the plugin and I had to do it. Unless I use rawStr workaround for extracting header and body in this PR. What should I do?

knocte: > Unless I use rawStr workaround for extracting headerStr and bodyStr in this PR. What should I do?

If the 17.4.0 PR is not merged yet, then you HAVE to use that workaround.

knocte:
@realmarv let's now rebase this PR 🙏

commitlint.config.ts
```diff
@@ -47,11 +47,15 @@ module.exports = {
         "too-many-spaces": [RuleConfigSeverity.Error, "always"],
         "commit-hash-alone": [RuleConfigSeverity.Error, "always"],
         "title-uppercase": [RuleConfigSeverity.Error, "always"],
+        "default-revert-message": [RuleConfigSeverity.Error, "never"],
```
knocte, line 50: @realmarv when I raised that we should use 'never' here instead of always, I also meant that the plugin should not ignore this argument, it should act on it (e.g. imagine if we contribute this rule at some point to commitlint repo and people want to use it with 'always' argument so that their devs can't edit the revert msg? I know it would be stupid but the point of this arg is to invert the logic)

realmarv: If someone edits the revert message they might edit it in a way that we can't even catch it's a revert message. We can check for revert keyword in the header and body to make sure if the commit is a revert but the rule's logic will change a bit.

commitlint/plugins.ts
realmarv: Done

commitlint/plugins.ts
realmarv: Done

commitlint/plugins.ts
realmarv: Done

commitlint.config.ts

commitlint/plugins.ts
realmarv: Done

commitlint.config.ts
realmarv: Done

realmarv:
> @realmarv let's now rebase this PR pray

Done

commitlint/plugins.ts
```diff
@@ -262,6 +262,44 @@ export abstract class Plugins {
...
+        headerStr: string,
+        bodyStr: string | null,
+        when = "never"
+    ) {
+        let offence = false;
+        let isRevertCommitMessage = headerStr.toLowerCase().includes("revert");
+        console.log("----");
+        console.log("headerStr: " + headerStr);
+        console.log("bodyStr:" + bodyStr);
+        console.log("----");
```
knocte, line 275: @realmarv let's remove debugging logs

commitlint/plugins.ts
```diff
@@ -262,6 +262,44 @@ export abstract class Plugins {
         ];
     }
 
+    public static defaultRevertMessage(
+        headerStr: string,
+        bodyStr: string | null,
+        when = "never"
```
knocte, line 268: @realmarv do we need a default value for when?

commitlint/plugins.ts
```diff
@@ -262,6 +262,44 @@ export abstract class Plugins {
...
+                }
+            }
+
+            const negated = when === "never";
+            offence = negated ? offence : !offence;
+        }
+        console.log("offence:" + offence);
+        return [
+            !offence,
+            `Please explain why you're reverting` + Helpers.errMessageSuffix,
```
knocte, line 299: this msg is only in the case of `never`, but we need a different message for `always`

commitlint/plugins.ts
```diff
@@ -262,6 +262,44 @@ export abstract class Plugins {
...
+                            null;
+                } else {
+                    offence = true;
+                }
+            }
+
+            const negated = when === "never";
+            offence = negated ? offence : !offence;
+        }
+        console.log("offence:" + offence);
```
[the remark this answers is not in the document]
realmarv: Done

commitlint/plugins.ts
realmarv: I was following the commitlint rule functions. Removed it

commitlint/plugins.ts
realmarv: Done

commitlint/plugins.ts
realmarv: Done

commitlint/plugins.ts
```diff
@@ -262,6 +262,44 @@ export abstract class Plugins {
...
+            }
+
+            offence = negated ? offence : !offence;
+        }
+        return [
+            !offence,
+            (negated
+                ? `Please explain why you're reverting`
+                : `Please don't change the default revert commit message`) +
+                Helpers.errMessageSuffix,
```
knocte, line 299: @realmarv if you check what is in Helpers.errMessageSuffix you will see that you need to end the previous error messages with a dot; otherwise it will read wrong

commitlint/plugins.ts
```diff
@@ -262,6 +262,44 @@ export abstract class Plugins {
...
+
+        if (isRevertCommitMessage) {
+            let isDefaultRevertHeader =
+                headerStr.match(/^[Rr]evert ".+"$/) !== null;
+
+            if (isDefaultRevertHeader) {
+                if (bodyStr !== null) {
+                    let lines = bodyStr.split("\n");
+                    offence =
+                        // 40 is the length of git commit hash.
```
knocte, line 283: @realmarv please move this comment down, there's no reason to not put the comment next to the line that is referring to

commitlint/plugins.ts
realmarv: Done

commitlint/plugins.ts
realmarv: Done

commitlint.config.ts
```diff
@@ -139,6 +143,32 @@ module.exports = {
...
+                    },
+                    when: string
+                ) => {
+                    let bodyStr = Helpers.convertAnyToString(body, "body");
+                    let headerUncastedStr = Helpers.convertAnyToString(
+                        header,
+                        "header"
+                    );
+                    let headerStr = Helpers.assertNotNull(
+                        headerUncastedStr,
```
knocte, line 162: @realmarv let's inline this variable because it's not being used anywhere else

commitlint/plugins.ts
```diff
@@ -262,6 +262,44 @@ export abstract class Plugins {
         ];
     }
 
+    public static defaultRevertMessage(
+        headerStr: string,
+        bodyStr: string | null,
+        when: string
```
knocte, line 268: @realmarv let's add an assert for this param, if it's not "never" or "always", it should throw an error

commitlint/plugins.ts
```diff
@@ -262,6 +262,44 @@ export abstract class Plugins {
...
+    ) {
+        let offence = false;
+        let isRevertCommitMessage = headerStr.toLowerCase().includes("revert");
```
knocte, line 271: @realmarv shouldn't it be "startsWith" instead of "includes"?

realmarv: If someone changes the commit title isn't it possible that "revert" word comes somewhere else in the title?

commitlint.config.ts
realmarv: What do you mean? Do you mean that I call Helpers.assertNotNull directly on the previous function call?

commitlint/plugins.ts
knocte: Yes, in which would mean that the revert message used is not default, so if when=never, then offence should be false in this case.

commitlint.config.ts
knocte: when there is a variable "let foo=1+2; someFunc(foo);" and I ask you to inline foo, then you simply doo "someFunc(1+2)"

commitlint/plugins.ts
realmarv: Done

commitlint.config.ts

commitlint/plugins.ts
realmarv: I added a test for this case

knocte: cool, but why do I still see the use of includes()?

realmarv: Because this variable is checking if the commit message is a revert commit message not if it's a default revert message

commitlint/helpers.ts
```diff
@@ -37,6 +37,14 @@ export abstract class Helpers {
         return text as string;
     }
 
+    public static assertWhen(when: string) {
+        if (when !== "never" && when !== "always") {
```
knocte, line 41: @realmarv I've realised we should create an enum called "When" that has both When.Never and When.Always (but instead of mapping to int, they map to string); hopefully typescript supports this?

commitlint/helpers.ts
```diff
@@ -38,7 +43,7 @@ export abstract class Helpers {
     }
 
     public static assertWhen(when: string) {
```
knocte, line 45: @realmarv you should change return type of this function to be When instead of void, in this commit (like we have done in other assert function); btw why is CI red?

realmarv: I don't know maybe some commitlint new version bug again

realmarv: > @realmarv you should change return type of this function to be When instead of void, in this commit

Why should I return When here?

knocte: > Why should I return When here?

To do the cast, like we did from string|null to string.

> I don't know maybe some commitlint new version bug again

Then this is the highest priority thing to fix.

realmarv: It's a syntax error in cosmiconfig module. I'm not sure if it's a bug in cosmiconfig or in typescript compiler

knocte: Well, it needs to be investigated.

knocte:
Let's rebase this PR to get rid of the red CI.

commitlint/helpers.ts
knocte: > To do the cast, like we did from string|null to string.

Let's not forget this.

commitlint.config.ts
```diff
@@ -151,7 +151,7 @@ module.exports = {
                         header: any;
                         body: any;
                     },
-                    when: string
+                    when: When
```
knocte, line 154: this change is wrong, the When type lives in our code, not in commitlint code

commitlint/helpers.ts
```diff
@@ -42,12 +42,13 @@ export abstract class Helpers {
         return text as string;
     }
 
-    public static assertWhen(when: string) {
+    public static assertWhen(when: When) {
```
knocte, line 45: @realmarv this change is wrong, when should be string, but the return type should be When (I told you to look at how we already do this in assertNotNull()

commitlint.config.ts
realmarv: fixed it

commitlint/helpers.ts
realmarv: fixed it

commitlint/helpers.ts
realmarv: Done

webwarrior-ws:
Superseded by https://github.com/nblockchain/conventions/pull/190

PR #29, opened 2022-11-10
F# scripts that detects .fsx files without shebang

realmarv:

PR #71, opened 2023-02-20
 F# scripts that detects .fsx files without shebang

realmarv:
Supersedes https://github.com/nblockchain/conventions/pull/29

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -114,3 +114,5 @@ jobs:
           # make git ignore mode changes in files and doesn't include them in the git diff command.
           git config core.fileMode false
           git diff --exit-code
+      - name: Run script to find files without shebang
+        run: dotnet fsi scripts/shebangConvention.fsx
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -3,7 +3,30 @@ name: CI
 on: [push, pull_request]
 
 jobs:
-  tests:
+  file-conventions:
+    runs-on: ubuntu-22.04
+    container:
+      image: "ubuntu:22.04"
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install required dependencies
+        run: |
+          apt update
+          apt install -y sudo
+          # We need to install curl otherwise we get these errors in the CI:
+          # Unable to load the service index for source https://api.nuget.org/v3/index.json.
+          # The SSL connection could not be established, see inner exception. 
+          # The remote certificate is invalid because of errors in the certificate chain: UntrustedRoot 
+          sudo apt install -y curl
+
+      - name: Setup .NET
+        run: apt install -y dotnet6
+      - name: compile the FileConventions project
+        run: dotnet build src/FileConventions.Test/FileConventions.Test.fsproj
+      - name: Run tests to validate F# scripts
+        run: dotnet test src/FileConventions.Test/FileConventions.Test.fsproj
+
+  commitlint-plugins-tests:
     name: Run commitlint-related tests
     runs-on: ubuntu-22.04
     container:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -78,8 +78,6 @@ jobs:
         run: sudo apt install -y dotnet6
       - name: Check all files end with EOL
         run: dotnet fsi scripts/eofConvention.fsx
-      - name: Check there are no mixed line-endings in any files
-        run: dotnet fsi scripts/mixedLineEndings.fsx
       - name: Install prettier
         run: npm install prettier@2.8.3
       - name: Change file permissions
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,30 @@
 .DS_Store
 node_modules/
 yarn.lock
+
+# Build results
+[Dd]ebug/
+[Dd]ebugPublic/
+[Rr]elease/
+[Rr]eleases/
+x64/
+x86/
+bld/
+[Bb]in/
+[Oo]bj/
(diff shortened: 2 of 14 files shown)


scripts/shebangConvention.fsx
```diff
@@ -0,0 +1,31 @@
...
+let invalidFiles = 
+    Directory.GetFiles(".", "*.fsx", SearchOption.AllDirectories) 
+    |> Seq.map (fun pathStr -> FileInfo pathStr)
+    |> Seq.filter (NotInDir "node_modules")
+    |> Seq.filter (NotInDir ".git")
+    |> Seq.filter (NotInDir "bin")
+    |> Seq.filter (NotInDir "obj")
+    |> Seq.filter (NotInDir "DummyFiles")
+    |> Seq.filter (fun fileInfo -> (not (FileConventions.HasCorrectShebang fileInfo) || 
+                                    not (FileConventions.IsExecutable fileInfo)))
```
knocte, line 21: @realmarv I'd rather you remove the call to FileConventions.IsExecutable from this script, because the script name only mentions shebang anyway (and then you can remove the step and the FIXME)

.github/workflows/CI.yml
```diff
@@ -21,6 +21,12 @@ jobs:
 
       - name: Setup .NET
         run: apt install -y dotnet6
+      - name: Set execute permissions for dummy files
+        run: |
+          chmod +x src/FileConventions.Test/DummyFiles/DummyExecutable.fsx
+          chmod -x src/FileConventions.Test/DummyFiles/DummyNotExecutable.fs
+          ls -l src/FileConventions.Test/DummyFiles/DummyExecutable.fsx
+          ls -l src/FileConventions.Test/DummyFiles/DummyNotExecutable.fs
```
knocte, line 29: @realmarv these ls calls are for debugging, I guess, so they should be removed

.github/workflows/CI.yml
```diff
@@ -21,6 +21,12 @@ jobs:
 
       - name: Setup .NET
         run: apt install -y dotnet6
+      - name: Set execute permissions for dummy files
+        run: |
+          chmod +x src/FileConventions.Test/DummyFiles/DummyExecutable.fsx
+          chmod -x src/FileConventions.Test/DummyFiles/DummyNotExecutable.fs
```
knocte, line 27: @realmarv rather do this in the tests please

.github/workflows/CI.yml
realmarv: Done

.github/workflows/CI.yml
knocte: why?

knocte:
> That is because it's not working properly and CI becomes red for no reason.

For no reason? You told me on telegram that it has a bug, therefore explain here what's the bug.

.github/workflows/CI.yml
realmarv: I'm adding execution permission to dummy files for FileConventions tests. We had a job called tests and I thought you meant I move running these tests to that job. Did you mean I do this chmods in the `Run tests to validate F# scripts` step?

knocte: Yes because I assume that those chmod calls are needed for the tests to pass. And if a developer decides to run the tests locally, they would fail if you don't place them within the tests, because a developer doesn't run CI locally to run tests.

knocte:
> GitHubCI: disable mixedLineEndings rule
> That is because MixedLineEndings function has a bug.
> If the line endings are "\r\n" it always returns true
> and CI becomes red.

That's much better explanation thanks. I'm gonna be afk soon, if you become taskless, you can create a PR that adds tests to mixedLineEndings in the same way this PR does (so that mixedLineEndings.fsx logic ends up in a library instead of being in the script itself).

.github/workflows/CI.yml
```diff
@@ -117,3 +117,10 @@ jobs:
           # make git ignore mode changes in files and doesn't include them in the git diff command.
           git config core.fileMode false
           git diff --exit-code
+      - name: Set execute permissions for fsx files
+        run: |
+          chmod +x scripts/eofConvention.fsx
+          chmod +x scripts/shebangConvention.fsx
+          chmod +x scripts/mixedLineEndings.fsx
```
knocte, line 124: @realmarv after you have removed the call to IsExecutable from shebangConventions.fsx, this step above is unnecesary

.github/workflows/CI.yml

.github/workflows/CI.yml

realmarv:
> > GitHubCI: disable mixedLineEndings rule
> > That is because MixedLineEndings function has a bug.
> > If the line endings are "\r\n" it always returns true
> > and CI becomes red.
> 
> That's much better explanation thanks. I'm gonna be afk soon, if you become taskless, you can create a PR that adds tests to mixedLineEndings in the same way this PR does (so that mixedLineEndings.fsx logic ends up in a library instead of being in the script itself).

Done

.github/workflows/CI.yml
knocte: This is not done yet, I still see some chmod calls in CI.yml

.gitignore
```diff
@@ -1,3 +1,30 @@
...
+# NUNIT
+*.VisualState.xml
+TestResult.xml
+TestResults.xml
+
+# Visual Studio cache files
+# files ending in .cache can be ignored
+*.[Cc]ache
+# but keep track of directories ending in .cache
+!*.[Cc]ache/
```
knocte, line 30: @realmarv this file has to be modified in its own commit, it's not related to the failing tests

PR #72, opened 2023-02-21
Add MixedLineEndings function to FileConventions

realmarv:


.gitignore
```diff
@@ -1,3 +1,30 @@
...
+# NUNIT
+*.VisualState.xml
+TestResult.xml
+TestResults.xml
+
+# Visual Studio cache files
+# files ending in .cache can be ignored
+*.[Cc]ache
+# but keep track of directories ending in .cache
+!*.[Cc]ache/
```
knocte, line 30: @realmarv this change should be in its own separate commit

src/FileConventions.Test/DummyFiles/DummyWithCRLFLineEndings
```diff
@@ -0,0 +1,3 @@
+line1
+line2
+line3
```
knocte, line 3: @realmarv this is confusing; what's the difference between this file and DummyWithoutMixedLineEndings file? because it seems but have line endings which are not mixed

src/FileConventions.Test/FileConventions.Test.fs
```diff
@@ -0,0 +1,28 @@
...
+    Assert.That(MixedLineEndings fileInfo, Is.EqualTo true)
+
+[<Test>]
+let MixedLineEndingsTest2 () =
+    let fileInfo = (FileInfo (Path.Combine(__SOURCE_DIRECTORY__, "DummyFiles", "DummyWithoutMixedLineEndings")))
+    Assert.That(MixedLineEndings fileInfo, Is.EqualTo false)
+
+[<Test>]
+let MixedLineEndingsTest3 () =
+    let fileInfo = (FileInfo (Path.Combine(__SOURCE_DIRECTORY__, "DummyFiles", "DummyWithCRLFLineEndings")))
```
knocte, line 27: @realmarv please save `Path.Combine(__SOURCE_DIRECTORY__, "DummyFiles"` in a DirectoryInfo variable to not have to repeat it in every test; this is a bad practice whose fix is usually called "DRY" (Don't Repeat Yourself), and I'd expect you from now on detect this duplication yourself (when you're about to commit, in the `git gui` view, it's very easy to spot duplication by looking at your own diff)

src/FileConventions.Test/DummyFiles/DummyWithCRLFLineEndings
knocte: @realmarv ^?

realmarv: I renamed DummyWithoutMixedLineEndings file. Is that ok?

src/FileConventions.Test/FileConventions.Test.fs
realmarv: I removed all duplicates but I'm not sure if was ok to remove duplicates of other tests as well

knocte:
I don't know what's going on with this PR, I see many weird things. Maybe it's better to discuss in Jitsi, but I'm going to start here with the first thing: why is the first commit called "add tests" (with green CI), instead of "add failing tests" (with red CI)?

src/FileConventions.Test/DummyFiles/DummyWithCRLFLineEndings
knocte: that sounds better, but why did you remove DummyWithCRLFLineEndings?

realmarv: DummyWithCRLFLineEndings is added in the third commit

realmarv:
> I don't know what's going on with this PR, I see many weird things. Maybe it's better to discuss in Jitsi, but I'm going to start here with the first thing: why is the first commit called "add tests" (with green CI), instead of "add failing tests" (with red CI)?

The first commit includes a failing test and another test that doesn't fail. Fixed the CI status

knocte:
Yesterday I was telling you (not sure if in this same PR) that you have to move the logic from script to library in the 1st commit, otherwise the fix cannot be seen in the diffs.

PR #74, opened 2023-02-23
Ignore binary files in eofConvention.fsx

realmarv:
Superseded https://github.com/nblockchain/conventions/pull/35


knocte:
Unfortunately this PR needs to be rebased.

scripts/eofConvention.fsx
```diff
@@ -4,22 +4,15 @@ open System.IO
...
-
-let EolAtEof(fileInfo: FileInfo) = 
-    use streamReader = new StreamReader (fileInfo.FullName)
-    let filetext = streamReader.ReadToEnd()
-    
-    if filetext <> String.Empty then
-        Seq.last filetext = '\n'
-    else
-        true
+let rootDir = Path.Combine(__SOURCE_DIRECTORY__, "..") |> DirectoryInfo
```
knocte, line 9: @realmarv nit: this line doesn't need to change

src/FileConventions.Test/FileConventions.Test.fs
```diff
@@ -78,3 +79,39 @@ let DetectAsteriskInPackageReferenceItems1 () =
...
+[<Test>]
+let HasBinaryContentTest2 () =
+    let fileInfo = (FileInfo (Path.Combine(__SOURCE_DIRECTORY__, "DummyFiles", "white.png")))
+    Assert.That(HasBinaryContent fileInfo, Is.EqualTo true)
+
+
+[<Test>]
+let HasBinaryContentTest3 () =
+    let fileInfo = (FileInfo (Path.Combine(__SOURCE_DIRECTORY__, "DummyFiles", "DummyNonBinaryFile.txt")))
+    Assert.That(HasBinaryContent fileInfo, Is.EqualTo false)
```
knocte, line 117: @realmarv tests should use common var dummyFilesDirectory

realmarv: Done

scripts/eofConvention.fsx
realmarv: Done

knocte:
This PR needs to be rebased and there are 2 comments unaddressed.

realmarv:
> This PR needs to be rebased and there are 2 comments unaddressed.

Rebase is done. What comments?

knocte:
The commit that says "Transfer the EolAtEof from eofConventions.fsx to FileConventions library." is not transferring the logic, it's changing the logic.

realmarv:
> The commit that says "Transfer the EolAtEof from eofConventions.fsx to FileConventions library." is not transferring the logic, it's changing the logic.

It's not changing the logic it's just using the enum you suggested instead of booleans

knocte:
> It's not changing the logic it's just using the enum you suggested instead of booleans

That's what I thought too, but no, you're adding one more else line.

realmarv:
> > It's not changing the logic it's just using the enum you suggested instead of booleans
> 
> That's what I thought too, but no, you're adding one more else line.

It's doing the same thing. More else lines is because the output is not boolean anymore

knocte:
> It's doing the same thing. More else lines is because the output is not boolean anymore

But there's no need to introduce that DU at that time, it's later. Let's take in account that when rebasing.

realmarv:
> > It's doing the same thing. More else lines is because the output is not boolean anymore
> 
> But there's no need to introduce that DU at that time, it's later. Let's take in account that when rebasing.

Done

.github/workflows/CI.yml
```diff
@@ -133,4 +133,5 @@ jobs:
           dotnet new tool-manifest
           dotnet tool install fantomless-tool --version 4.7.997-prerelease
           dotnet fantomless --recurse .
+          git restore src/FileConventions.Test/DummyFiles
```
knocte, line 136: @realmarv the test files for EolAtEof don't need to have .fsx extension; this way we don't need this `git restore` call at all

PR #85, opened 2023-03-11
Detect inconsistent pulumi versions in GitHubCI

realmarv:


src/FileConventions/Library.fs
```diff
@@ -104,3 +103,37 @@ let EolAtEof(fileInfo: FileInfo) =
                 False
         else
             True
+
+let DetectInconsistentVersionsInGitHubCI(fileInfos: seq<FileInfo>) =
```
knocte, line 107: @realmarv good work! now let's rename this func to DetectInconsistentVersionsInGitHubCIWorkflow, then create a new func DetectInconsistentVersionsInGitHubCI that receives a DirectoryInfo instead of FileInfo, which will recurively get all files underneath the directory and call DetectInconsistentVersionsInGitHubCIWorkflow with a list of all of them

knocte:
I think commit 9caf should be squashed with commit 871

realmarv:
This PR is closed and superseded by https://github.com/nblockchain/conventions/pull/85 because some commits needed to be squashed.

Issue #82, opened 2023-03-08
Check if gitPush1by1 was used


knocte:
The way to do it:
- For a PR (so, a CI step that has an `if: github.event_name == 'pull_request'`), a CI job should be triggered that checks if all commits in the PR have a CI status, using:
    - This API: https://docs.github.com/en/rest/checks/suites
    - The commits info: `${{ github.event.pull_request.head.sha }}` and `${{ github.event.pull_request.commits }}`

PR #88, opened 2023-03-14
Add a script to check if gitPush1by1 was used

realmarv:
Fixes https://github.com/nblockchain/conventions/issues/82

### DIFF
--- a/scripts/detectNotUsingGitPush1by1.fsx
+++ b/scripts/detectNotUsingGitPush1by1.fsx
@@ -1,0 +1,73 @@
+#!/usr/bin/env -S dotnet fsi
+
+open System
+open System.Net.Http
+open System.Net.Http.Headers
+
+#r "nuget: Fsdk, Version=0.6.0--date20230214-0422.git-1ea6f62"
+
+let gitRepo = Environment.GetEnvironmentVariable "GITHUB_REPOSITORY"
+
+if gitRepo = null then
+    failwith "You shouldn't use this script outside CI."
+
+let currentBranch =
+    Fsdk
+        .Process
+        .Execute(
+            {
+                Command = "git"
+                Arguments = "rev-parse --abbrev-ref HEAD"
+            },
+            Fsdk.Process.Echo.Off
+        )
+        .UnwrapDefault()
+        .Trim()
+
+let prCommits =
+    Fsdk
+        .Process
+        .Execute(
+            {
+                Command = "git"
+                Arguments =
+                    sprintf "rev-list %s~..%s" currentBranch currentBranch
+            },
+            Fsdk.Process.Echo.Off
+        )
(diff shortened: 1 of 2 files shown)


scripts/detectNotUsingGitPush1by1.fsx
```diff
@@ -0,0 +1,65 @@
...
+        )
+        .UnwrapDefault()
+        .Trim()
+        .Split "\n"
+    |> Seq.tail
+
+let notUsingGitPush1by1 =
+    prCommits
+    |> Seq.map(fun commit ->
+        let client: HttpClient = new HttpClient()
```
knocte, line 41: @realmarv is client IDisposable? if yes, then it should use `use`, not `let`; if not, then it shouldn't use `new` keyword (this comment applies to other line where you're also using `new`)

knocte:
Two things:
- How is it possible that you didn't need ${{ github.event.pull_request.commits }} ? Are you sure the logic is correct?
- You forgot the "Fixes $URL" at the end.

knocte:
And CI is red actually 😬

realmarv:
> Two things:
> 
> * How is it possible that you didn't need ${{ github.event.pull_request.commits }} ? Are you sure the logic is correct?
> * You forgot the "Fixes $URL" at the end.

`git rev-list HEAD~..HEAD` command returns the PR commits in pull requests. (It returns only the last commit when I try it locally, but in PR, it returns PR commits.)

scripts/detectNotUsingGitPush1by1.fsx

scripts/detectNotUsingGitPush1by1.fsx
```diff
@@ -0,0 +1,64 @@
...
+
+let currentBranch =
+    Fsdk
+        .Process
+        .Execute(
+            {
+                Command = "git"
+                Arguments = "rev-parse --abbrev-ref HEAD"
+            },
+            Fsdk.Process.Echo.All
```
knocte, line 17: @realmarv let's use Echo.Off

scripts/detectNotUsingGitPush1by1.fsx
```diff
@@ -0,0 +1,64 @@
...
+let prCommits =
+    Fsdk
+        .Process
+        .Execute(
+            {
+                Command = "git"
+                Arguments =
+                    sprintf "rev-list %s~..%s" currentBranch currentBranch
+            },
+            Fsdk.Process.Echo.All
```
knocte, line 31: @realmarv let's use Echo.Off

knocte:
BTW I pushed 2 WIP commits to test if the build fails, and it does! good work. You can remove those two WIP commits.

scripts/detectNotUsingGitPush1by1.fsx

PR #91, opened 2023-03-23
Detect non-verbose flags being used in F# scripts or YML CI files

realmarv:


knocte:
Please rebase this PR 🙏 Thanks

realmarv:
> Please rebase this PR pray Thanks

Done

src/FileConventions/Library.fs
```diff
@@ -104,3 +104,39 @@ let EolAtEof(fileInfo: FileInfo) =
                 False
         else
             True
+
+let allowedNonVerboseFlags = seq { "env -S" }
```
knocte, line 108: @realmarv we need to add "unzip" here, because it doesn't have long flags

knocte: (let's do it with a new unit test too)

realmarv: Done and rebased

src/FileConventions/Library.fs
```diff
@@ -329,3 +329,46 @@ let DetectInconsistentVersionsInFSharpScripts
...
+            ".fs"
+            ".sh"
+        }
+
+    assert
+        validExtensions
+        |> Seq.map(fun ext -> fileInfo.FullName.EndsWith(ext))
+        |> Seq.contains true
+
+    let fileLines = File.ReadLines(fileInfo.FullName)
```
knocte, line 356: @realmarv nit: no need for parentheses in lines 353 and 356

src/FileConventions/Library.fs
```diff
@@ -329,3 +329,46 @@ let DetectInconsistentVersionsInFSharpScripts
...
+            ".yml"
+            ".fsx"
+            ".fs"
+            ".sh"
+        }
+
+    assert
+        validExtensions
+        |> Seq.map(fun ext -> fileInfo.FullName.EndsWith(ext))
+        |> Seq.contains true
```
knocte, line 354: @realmarv can you explain what exactly are you doing here please? this is a bit unreadable; also, why are you using assert instead of failwith? I'm actually not familiar with `assert` func TBH, but I have the feeling that it only works in DEBUG mode and the error thrown would be very unreadable (as opposed to using failwith where you are required to provide a readable error)

knocte:
@realmarv I raised 2 nits before this can be merged; but please before addressing them, please pull, because I added a commit to your branch

knocte:
@realmarv also PR's CI is red

src/FileConventions/Library.fs
```diff
@@ -329,3 +329,46 @@ let DetectInconsistentVersionsInFSharpScripts
...
+let allowedNonVerboseFlags =
+    seq {
+        "unzip"
+
+        // even if env in linux has --split-string=foo as equivalent to env -S, it
+        // doesn't seem to be present in macOS' env man page and doesn't work either
+        "env -S"
+    }
+
+let NonVerboseFlagsInGitHubCI(fileInfo: FileInfo) =
```
knocte, line 342: @realmarv this function should be renamed; if we're checking for non-verbose flags in .fsx files then it has nothing to do with GitHubCI

src/FileConventions/Library.fs
knocte: @realmarv BTW I just cooked this in case you want to use it: https://github.com/nblockchain/fsx/commit/7acb7a5b8fa374ee2b68c28c2e438cb6c9265e58 (will be available in nuget in 5min)

src/FileConventions/Library.fs

src/FileConventions/Library.fs

src/FileConventions/Library.fs
realmarv: It loops over validExtensions sequence and checks if the file path ends with any of them

scripts/nonVerboseFlagsInGitHubCIAndScripts.fsx
```diff
@@ -0,0 +1,31 @@
...
+        ".fs"
+        ".sh"
+    }
+
+let invalidFiles =
+    validExtensions
+    |> Seq.map(fun ext ->
+        Helpers.GetInvalidFiles
+            rootDir
+            ("*" + ext)
```
knocte, line 24: @realmarv aren't you missing a dot after * here?

src/FileConventions/Library.fs
```diff
@@ -329,3 +329,50 @@ let DetectInconsistentVersionsInFSharpScripts
...
+        }
+
+    let isFileExtentionValid =
+        validExtensions
+        |> Seq.map(fun ext -> fileInfo.FullName.EndsWith ext)
+        |> Seq.contains true
+
+    if not isFileExtentionValid then
+        failwith
+            "NonVerboseFlags function only supports '.yml', '.fsx', '.fs' and '.sh' files."
```
knocte, line 358: @realmarv please respect DRY principle here

scripts/nonVerboseFlagsInGitHubCIAndScripts.fsx
realmarv: No ext already has a dot

knocte:
CI is still red

realmarv:
> CI is still red

It's green now

Issue #95, opened 2023-03-26
Improve scripts/detectNotUsingGitPush1by1.fsx to detect broken(red) CI statuses


knocte:
After detectNotUsingGitPush1by1.fsx checks that gitPush1by1 was used it should check CI status of each commit:
- If commit msg contains text "failing test", then it would be allowed to have broken CI.
- Else: CI should be green (and if it's not, then script should exit with ExitCode>0).

realmarv:
I'm not sure if it's possible because while the script is running in the CI, the CI status is in_progress. Unless we force push CI runs to finish before starting pull_request CI runs (Not sure if it's possible either)

knocte:
Let's make it for now that if it's red, it fails. If It's green or **in progress**, then it works.

PR #92, opened 2023-03-23
scripts/detectNotUsingGitPush1by1.fsx: fix

tehraninasab:
The script should query the fork repository that the PR
originated from, otherwise some commits might not have
CI status and this script will throw an error, while the
user has pushed the commits one by one.
The `GITHUB_EVENT_PATH` environment variable points to a JSON
file that contains information about the pull request event,
including the user and repository information of the fork
repository.

This is an example of the JSON file:

```
{
    "action": "opened",
    "number": 4,
    "pull_request": {
      "_links": {
        "comments": {
          "href": "https://api.github.com/repos/realmarv/conventions/issues/4/comments"
        },
        "commits": {
          "href": "https://api.github.com/repos/realmarv/conventions/pulls/4/commits"
        },
        "html": {
          "href": "https://github.com/realmarv/conventions/pull/4"
        },
        "issue": {
          "href": "https://api.github.com/repos/realmarv/conventions/issues/4"
        },
        "review_comment": {
          "href": "[https://api.github.com/repos/realmarv/conventions/pulls/comments{/number}](https://api.github.com/repos/realmarv/conventions/pulls/comments%7B/number%7D)"
        },
        "review_comments": {
          "href": "https://api.github.com/repos/realmarv/conventions/pulls/4/comments"
        },
        "self": {
          "href": "https://api.github.com/repos/realmarv/conventions/pulls/4"
        },
        "statuses": {
          "href": "https://api.github.com/repos/realmarv/conventions/statuses/c5eb50ea6c5a72d61ab85d96ffc66cc4741d3133"
        }
      },
      "active_lock_reason": null,
      "additions": 21,
      "assignee": null,
      "assignees": [],
      "author_association": "OWNER",
      "auto_merge": null,
      "base": {
        "label": "realmarv:master",
        "ref": "master",
        "repo": {
          "allow_auto_merge": false,
          "allow_forking": true,
          "allow_merge_commit": true,
          "allow_rebase_merge": true,
          "allow_squash_merge": true,
          "allow_update_branch": false,
          "archive_url": "[https://api.github.com/repos/realmarv/conventions/{archive_format}{/ref}](https://api.github.com/repos/realmarv/conventions/%7Barchive_format%7D%7B/ref%7D)",
          "archived": false,
          "assignees_url": "[https://api.github.com/repos/realmarv/conventions/assignees{/user}](https://api.github.com/repos/realmarv/conventions/assignees%7B/user%7D)",
[...]

PR #97, opened 2023-03-27
scripts/checkCommits1by1.fsx: augment

realmarv:
Improve this fsx script to also detect commits with bad CI
status. That is, if commit msg contains text "failing test"
then it would be allowed to have broken CI, else CI should be
green.

Fixes https://github.com/nblockchain/conventions/issues/95

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -179,9 +179,9 @@ jobs:
         run: dotnet fsi scripts/unpinnedNugetPackageReferenceVersions.fsx
       - name: Check there are no unpinned versions in `dotnet tool install` commands
         run: dotnet fsi scripts/unpinnedDotnetToolInstallVersions.fsx
-      - name: Check if gitPush1by1 was used
+      - name: Check commits 1 by 1
         if: github.event_name == 'pull_request'
-        run: dotnet fsi scripts/detectNotUsingGitPush1by1.fsx
+        run: dotnet fsi scripts/checkCommits1by1.fsx
       - name: Check there are no inconsistent versions GitHubCI files
         run: dotnet fsi scripts/inconsistentVersionsInGitHubCI.fsx
       - name: Check there are no inconsistent versions in nuget package references of F# scripts
--- a/scripts/checkCommits1by1.fsx
+++ b/scripts/checkCommits1by1.fsx
@@ -566,6 +566,8 @@ let gitForkUser = parsedJsonObj.PullRequest.Head.User.Login
 let gitForkRepo = parsedJsonObj.PullRequest.Head.Repo.Name
 let gitRepo = $"{gitForkUser}/{gitForkRepo}"
 
+Console.WriteLine $"Fork repository is: {gitRepo}"
+
 type PRCommitsType =
     JsonProvider<"""
 [
@@ -879,6 +881,9 @@ let prCommits =
     let parsedPrCommitsJsonObj = PRCommitsType.Parse prCommitsJsonString
     parsedPrCommitsJsonObj |> Seq.map(fun commit -> commit.Sha)
 
+Console.WriteLine
+    $"Pull request commits are: {Environment.NewLine}{String.concat Environment.NewLine prCommits}"
+
 let hasCiStatus =
     prCommits
     |> Seq.map(fun commit ->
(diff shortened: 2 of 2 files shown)


knocte:
> Also "Compile F# scripts" step in GitHubCI is
removed because there is no error in
detectNotUsingGitPush1by1.fsx script
but compileFSharpScripts.fsx script can not
compile this file.

WHAT????

First, I don't understand why it would not be able to compile this file. And second, don't you think you should explain why it cannot compile this file? The word *why* in WorkflowGuidelines.md is referring to the ultimate why. Do not hide information to the next developer.

knocte:
> First, I don't understand why it would not be able to compile this file. And second, don't you think you should explain why it cannot compile this file? The word why in WorkflowGuidelines.md is referring to the ultimate why. Do not hide information to the next developer.

Third, do you think it would be acceptable that we remove the ability to check the correctness of ALL F# SCRIPTS just because one of them cannot be compiled?

realmarv:
> > Also "Compile F# scripts" step in GitHubCI is
> > removed because there is no error in
> > detectNotUsingGitPush1by1.fsx script
> > but compileFSharpScripts.fsx script can not
> > compile this file.
> 
> WHAT????
> 
> First, I don't understand why it would not be able to compile this file. And second, don't you think you should explain why it cannot compile this file? The word _**why**_ in WorkflowGuidelines.md is referring to the ultimate why. Do not hide information to the next developer.

I think it should be investigated and resolved in fsxc. What can I do about it in this PR?

knocte:
> I think it should be investigated and resolved in fsxc

If you think it's a fsxc bug, then:
1) State so in the commit message, there's no reason to hide your suspicions.
2) Or even go file a bug in fsx repo, and then reference the bug URL in your commit message so that the next developer understands why the script cannot be compiled, by reading your commit message.

However, this problem you're having is basically the same we're dealing with in https://github.com/nblockchain/conventions/pull/92 , so fix PR92 first with my guidance there (by doing the thing I initially told you to), and then after you do that, you would be able to apply the same thing here.

By the way when I say **by doing the thing I initially told you to** I don't intend to sound snarky, but think about it from my point of view: I tell you to do A, then you decide that for doing A you would need to do something that looks a bit "ugly" (putting a very long string in an F# file) so you try to fix the ugliness by trying other technique B slightly different than A: you suffer problems with technique B, and decide to solve the problem with a workaround C that deep down you know would be unacceptable by me (eliminating an entire step from CI?), instead of forgetting about B and really doing A as I asked you to.

realmarv:
> > I think it should be investigated and resolved in fsxc
> 
> If you think it's a fsxc bug, then:
> 
> 1. State so in the commit message, there's no reason to hide your suspicions.
> 2. Or even go file a bug in fsx repo, and then reference the bug URL in your commit message so that the next developer understands why the script cannot be compiled, by reading your commit message.
> 
> However, this problem you're having is basically the same we're dealing with in #92 , so fix PR92 first with my guidance there (by doing the thing I initially told you to), and then after you do that, you would be able to apply the same thing here.
> 
> By the way when I say **by doing the thing I initially told you to** I don't intend to sound snarky, but think about it from my point of view: I tell you to do A, then you decide that for doing A you would need to do something that looks a bit "ugly" (putting a very long string in an F# file) so you try to fix the ugliness by trying other technique B slightly different than A: you suffer problems with technique B, and decide to solve the problem with a workaround C that deep down you know would be unacceptable by me (eliminating an entire step from CI?), instead of forgetting about B and really doing A as I asked you to.

I didn't know the compile error is because of ResolutionFolder. I thought when you say the code could be greatly simplified by using FSharp.Data JSON parser instead of regex, it's not ok to put 500 lines of JSON string in the script and it's supposed to become shorter. Sorry.

knocte:
Please rebase this PR 🙏 Thanks

knocte:
> I didn't know the compile error is because of ResolutionFolder.

I didn't know either, but you don't need to know; remember what I always say? Make it work, then after that, make it pretty. You were making it pretty (because 500 in-line string is _ugly_) before making it work.

scripts/detectNotUsingGitPush1by1.fsx
```diff
@@ -563,37 +563,441 @@ let prCommits =
...
-let notUsingGitPush1by1 =
-    prCommits
-    |> Seq.map(fun commit ->
-        use client = new HttpClient()
-        client.DefaultRequestHeaders.Accept.Clear()
-
-        client.DefaultRequestHeaders.Accept.Add(
-            MediaTypeWithQualityHeaderValue "application/vnd.github+json"
-        )
+type checkSuitsType =
```
knocte, line 566: @realmarv typo ^ also: types should be PascalCase

scripts/detectNotUsingGitPush1by1.fsx
```diff
@@ -563,37 +563,441 @@ let prCommits =
...
+        ).[0]
+
+    let commitHash = checkSuitsParsedJson.CheckSuites.[0].HeadSha
+    let status = checkSuitsParsedJson.CheckSuites.[0].Status
+    let conclusion = checkSuitsParsedJson.CheckSuites.[0].Conclusion
+
+    if status = "completed" then
+        if commitMessageTitle.Contains "failing test" && conclusion = "success" then
+            Console.Error.WriteLine
+                $"The CI status of the commit \"{commitHash}\" must be `failure`, since it's adding a failing test."
```
knocte, line 994: @realmarv I never asked you to do this (next time you think of new requirements of a task, ask me first please, as it's already explained in the handbook*); in fact it is wrong, because the commit msg could say "fix failing test"

scripts/detectNotUsingGitPush1by1.fsx
```diff
@@ -563,37 +563,441 @@ let prCommits =
...
+
+            Environment.Exit 2
+
+    else if conclusion = "failure" then
```
knocte, line 998: @realmarv use elif, not else if

scripts/detectNotUsingGitPush1by1.fsx
```diff
@@ -563,37 +563,441 @@ let prCommits =
...
+            .Split(
+            "\n"
+        ).[0]
+
+    let commitHash = checkSuitsParsedJson.CheckSuites.[0].HeadSha
+    let status = checkSuitsParsedJson.CheckSuites.[0].Status
+    let conclusion = checkSuitsParsedJson.CheckSuites.[0].Conclusion
+
+    if status = "completed" then
+        if commitMessageTitle.Contains "failing test" && conclusion = "success" then
```
knocte, line 992: @realmarv I think these nested if clauses are a bit unreadable, please use a multi-match line instead, i.e. beginning with `match status, conclusion with` (the Contains call can be then put with a `when`)

realmarv: I guess there is no need to use nested if clauses; only one `if` is enough. Should I still use `match status, conclusion with`?

scripts/detectNotUsingGitPush1by1.fsx
realmarv: Removed it

scripts/detectNotUsingGitPush1by1.fsx

knocte: this typo is not addressed, read the message again: I'm pointing out 2 things, not 1

scripts/detectNotUsingGitPush1by1.fsx
```diff
@@ -563,37 +563,437 @@ let prCommits =
...
-let notUsingGitPush1by1 =
-    prCommits
-    |> Seq.map(fun commit ->
-        use client = new HttpClient()
-        client.DefaultRequestHeaders.Accept.Clear()
-
-        client.DefaultRequestHeaders.Accept.Add(
-            MediaTypeWithQualityHeaderValue "application/vnd.github+json"
-        )
+type CheckSuitsType =
```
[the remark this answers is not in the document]
realmarv: Ah right

knocte:
Unfortunately there are conflicts, please do a rebase when you have time (low-priority task).

knocte:
@realmarv can you rebase this PR please

scripts/detectNotUsingGitPush1by1.fsx
```diff
@@ -842,3 +844,416 @@ if notUsedGitPush1by1 then
...
+
+    let status = checkSuitesParsedJson.CheckSuites.[0].Status
+    let conclusion = checkSuitesParsedJson.CheckSuites.[0].Conclusion
+
+    if
+        status = "completed" && conclusion = "failure"
+        && not(commitMessageTitle.Contains "failing test")
+    then
+        Console.Error.WriteLine
+            $"The CI status of a commit that doesn't add a failing test must be `success`."
```
knocte, line 1256: @realmarv nit: if you're not interpolating anything, there's no need for $ char. Anyway let's improve this message, I think it's better written like this: "Thanks for pushing commits 1 by 1, however we have detected some of them to not be successful (or not be red when they add a failing test)", which makes me think that this PR should also detect if a commit msg contains "failing test" and CI status is green, it should fail as well

knocte:
This PR needs to be rebased, when you have time. I've also realised that since the script would gain a new feature (not related to "detect not using gitPush1by1"), then we should rename the script. Maybe `analyzeCommits1by1.fsx`?

scripts/detectNotUsingGitPush1by1.fsx
```diff
@@ -842,3 +847,413 @@ if notUsedGitPush1by1 then
...
+
+    let status = checkSuitesParsedJson.CheckSuites.[0].Status
+    let conclusion = checkSuitesParsedJson.CheckSuites.[0].Conclusion
+
+    if
+        status = "completed" && conclusion = "failure"
+        && not(commitMessage.Contains "failing test")
+    then
+        Console.Error.WriteLine
+            "Thanks for pushing commits 1 by 1, however we have detected some of them to not be successful (or not be red when they add a failing test)"
```
knocte, line 1256: @realmarv nit: `we have detected some` -> `it has been detected that some` (as using "we" is a bit weird)

PR #103, opened 2023-04-04
Add scripts to apply and check the style of F#, C#, TS, YMl and XAML files

tehraninasab:
Add styleChecker.fsx script to check style of our F#, TS
and YML codes.

The `git respore package.json` command in the 
styleChecker.fsx script was failing with the following
error, even when I was running the 
`git config --global --add safe.directory '*'` command in
both the styleChecker.fsx script and in the CI. In the 
issue [1], it's suggested by someone to use --system
instead of --global in the mentioned command, when the 
git command is running in a container, which solved the 
problem.

```
fatal: detected dubious ownership in repository at '/__w/conventions/conventions'
To add an exception for this directory, call:

        git config --global --add safe.directory /__w/conventions/conventions

Error when running 'git restore package.json'
Fsdk.Process+ProcessFailed: Exception of type 'Fsdk.Process+ProcessFailed' was thrown.
   at Fsdk.Process.ProcessResult.Unwrap(String errMsg)
   at Fsdk.Process.ProcessResult.UnwrapDefault()
   at FSI_0002.RunPrettier(String arguments)
   at <StartupCode$FSI_0002>.$FSI_0002.main@()
Stopped due to error
Error: Process completed with exit code 1.
```

[1] https://github.com/actions/checkout/issues/1048


scripts/styleChecker.fsx
```diff
@@ -0,0 +1,99 @@
...
+                    Command = "npx"
+                    Arguments = $"prettier {arguments}"
+                },
+                Process.Echo.All
+            )
+            .Result
+
+    match processResult with
+    | ProcessResultState.Success output -> output
+    | ProcessResultState.WarningsOrAmbiguous output -> output.StdErr
```
knocte, line 57: @realmarv in this case you're printing StdErr but not StdOut

knocte (changes requested):

knocte:
Please change GitHubCI to use this script also.

scripts/styleChecker.fsx
```diff
@@ -0,0 +1,99 @@
...
+                    Command = "npx"
+                    Arguments = $"prettier {arguments}"
+                },
+                Process.Echo.All
+            )
+            .Result
+
+    match processResult with
+    | ProcessResultState.Success output -> output
+    | ProcessResultState.WarningsOrAmbiguous output -> output.StdOut
```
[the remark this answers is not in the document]
realmarv: I used the Unwrap method approach for printing the output. Not sure if it's ok

realmarv:
> Please change GitHubCI to use this script also.

Done

scripts/styleChecker.fsx
```diff
@@ -0,0 +1,149 @@
...
+
+let suggestion =
+    "Please use the following commands to style your code:"
+    + System.Environment.NewLine
+    + "Style your F# code using: `dotnet fantomless --recurse .`"
+    + System.Environment.NewLine
+    + "Style your TypeScript code using: `npx prettier --quote-props=consistent --write ./**/*.ts`"
+    + System.Environment.NewLine
+    + "Style your YML code using: `npx prettier --quote-props=consistent --write ./**/*.yml`"
+    + System.Environment.NewLine
```
knocte, line 128: @realmarv you could run git diff each time after you run each command, so that the error message is more specific and not so big

.github/workflows/CI.yml
```diff
@@ -146,24 +146,6 @@ jobs:
...
-          git restore package.json
-          git diff --exit-code
-      - name: fantomless
-        run: |
-          dotnet new tool-manifest
-          dotnet tool install fantomless-tool --version 4.7.997-prerelease
-          dotnet fantomless --recurse .
-          git diff --exit-code
+      - name: Check style of our F#, TypeScript and YML code
+        run: sudo dotnet fsi scripts/styleChecker.fsx
```
knocte, line 150: @realmarv please make the `npm install prettier@2.8.3` also part of the script, but only run it if you detect there are .ts or .yml files in the repo (I know there are in this repo, but maybe in others where we will run this script, there won't be, so there's no need to install prettier in that case)

scripts/styleChecker.fsx
realmarv: Done

.github/workflows/CI.yml
realmarv: Done

scripts/styleChecker.fsx
```diff
@@ -0,0 +1,99 @@
...
+
+Process
+    .Execute(
+        {
+            Command = "git"
+            Arguments = "diff --exit-code"
+        },
+        Process.Echo.Off
+    )
+    .UnwrapDefault()
```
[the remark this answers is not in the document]
realmarv: Done

scripts/styleChecker.fsx
realmarv: Done

knocte:
> In the [1] issue

I guess you meant `in the issue [1]` ???

scripts/styleChecker.fsx
```diff
@@ -0,0 +1,266 @@
...
+        )
+        .UnwrapDefault()
+    |> ignore
+
+    Process
+        .Execute(
+            {
+                Command = "dotnet"
+                Arguments =
+                    "tool install fantomless-tool --version 4.7.997-prerelease"
```
knocte, line 33: @realmarv let's put this version at the top to make it easy to change, in the same way you did with the prettier version

realmarv:
> > In the [1] issue
> 
> I guess you meant `in the issue [1]` ???

Right

scripts/styleChecker.fsx
```diff
@@ -0,0 +1,266 @@
...
+    | WarningsOrAmbiguous output ->
+        if processResult.Details.Echo = Echo.Off then
+            output.PrintToConsole()
+            Console.WriteLine()
+            Console.Out.Flush()
+
+        let fullErrMsg = sprintf "%s (with warnings?)" errMsg
+        fullErrMsg
+
+    |> printfn "%A"
```
knocte, line 193: @realmarv if an error message is being printed to the console, then Console.Error.WriteLine  should be used, not prinfn (which is equivalent to Console.WriteLine)

knocte (changes requested):

scripts/styleChecker.fsx
realmarv: Done

scripts/styleChecker.fsx
```diff
@@ -0,0 +1,263 @@
...
+
+let CheckStyleOfYmlFiles(rootDir: DirectoryInfo) : int =
+    let suggestion =
+        "Please style your YML code using: `npx prettier --quote-props=consistent --write ./**/*.yml`"
+
+    GitRestore()
+
+    let exitCode =
+        if ContainsFiles rootDir "*.yml" then
+            InstallPrettier(prettierVersion)
```
knocte, line 242: @realmarv no need for parens here

scripts/styleChecker.fsx
```diff
@@ -0,0 +1,263 @@
...
+            StyleYmlFiles()
+            let processResult = GitDiff()
+            PrintProcessResult processResult suggestion
+            GetProcessExitCode processResult
+        else
+            0
```
knocte, line 248: @realmarv hardcoding 0 here is a bit weird, why dont call the var `success` instead of `exitCode`, and then here you hardcode `true`

scripts/styleChecker.fsx
realmarv: Done

scripts/styleChecker.fsx
realmarv: Done

scripts/styleChecker.fsx
```diff
@@ -0,0 +1,263 @@
...
+            Console.WriteLine()
+            Console.Out.Flush()
+
+        let fullErrMsg = sprintf "%s (with warnings?)" errMsg
+        Console.Error.WriteLine fullErrMsg
+
+let GetProcessExitCode(processResult: ProcessResult) : int =
+    match processResult.Result with
+    | Success output -> 0
+    | _ -> 1
```
knocte, line 195: @realmarv same here about hardcoding this, you can rename func to IsProcessSuccessful, and return false here

scripts/styleChecker.fsx
```diff
@@ -0,0 +1,393 @@
...
+    |> ignore
+
+    Process.Execute(
+        {
+            Command = "npx"
+            Arguments = $"prettier {arguments}"
+        },
+        Echo.Off
+    )
+    |> UnwrapProcessResult "" true
```
knocte, line 226: @realmarv when you are calling UnwrapProcessResult with `""` and `true`, does it act differently than Process.UnwrapResult function from Fsdk?

realmarv: when the process result type is WarningsOrAmbiguous, ProcessResult.Unwrap function raises an error but we don't want to raise that error because if you remember, npm had an unexpected output (I guess it was writing its output in stderr even when there was no error or something)

knocte: @realmarv so, you're saying that when passing `true` to UnwrapProcessResult (which means that the param `raiseError` = true), then you DO NOT raise an error (if warning)? If the answer to this is yes, do you understand why this is extremely confusing to me?

scripts/styleChecker.fsx
```diff
@@ -0,0 +1,393 @@
...
+    |> ignore
+
+let RunPrettier(arguments: string) =
+
+    // We need this step so we can change the files using `npx prettier --write` in the next step.
+    // Otherwise we get permission denied error in the CI.
+    Process
+        .Execute(
+            {
+                Command = "chmod"
```
knocte, line 211: @realmarv let's only do this in case environment variable GITHUB_REF is not empty (to make sure this is only done when in CI; take in account that we may ask our teammates to run this script soon from their local environments, not just CI)

scripts/styleChecker.fsx
```diff
@@ -0,0 +1,393 @@
...
+                    Arguments = $"install @prettier/plugin-xml@{version}"
+                },
+                Echo.Off
+            )
+            .UnwrapDefault()
+        |> ignore
+
+let StyleXamlFiles() =
+    InstallPrettier(prettierVersion)
+    InstallPrettierPluginXml(pluginXmlVersion)
```
knocte, line 178: @realmarv nit, no need for parentheses in the 2 lines above

scripts/styleChecker.fsx
```diff
@@ -0,0 +1,393 @@
...
+                Command = "npm"
+                Arguments = $"install prettier@{version}"
+            },
+            Echo.Off
+        )
+        |> UnwrapProcessResult "" true
+        |> ignore
+
+let StyleFSharpFiles(rootDir: DirectoryInfo) =
+    InstallFantomlessTool(fantomlessToolVersion)
```
knocte, line 128: @realmarv nit, no need for parentheses in the 2 lines above

scripts/styleChecker.fsx
```diff
@@ -0,0 +1,393 @@
...
+            },
+            Echo.Off
+        )
+        .UnwrapDefault()
+    |> ignore
+
+    Process
+        .Execute(
+            {
+                Command = "./node_modules/.bin/prettier"
```
knocte, line 195: @realmarv hardcoding forward-slashes in an F# file is a bad practice (you should use `Path.Combine` for that), in fact here if you do this, it may work in Linux but fail in Windows

scripts/styleChecker.fsx
```diff
@@ -0,0 +1,393 @@
...
+    Helpers.GetFiles rootDir searchPattern |> Seq.length > 0
+
+let GitDiff() : ProcessResult =
+
+    // Since we changed file modes in the prettier step we need the following command to
+    // make git ignore mode changes in files and doesn't include them in the git diff command.
+    Process
+        .Execute(
+            {
+                Command = "git"
```
knocte, line 259: @realmarv let's do this also only when run in CI (so check for github env var)

knocte: @realmarv actually this whole function shouldn't be called if not in CI

knocte:
@realmarv also please rebase this PR when you have time (BTW the reason why I thought this PR might become kinda important soon is that yesterday I saw the guys arguing over having too many EOLs in NX repo, like 2 EOLs instead of 1 EOL, which I think it's ridiculous bike-shedding, which can only be fixed by enforcing fantomless, so that those conversations are not needed at all ever again)

.github/workflows/CI.yml
```diff
@@ -144,26 +144,5 @@ jobs:
...
-          # run the following command to ignore package.json file
-          git restore package.json
-          git diff --exit-code
-      - name: fantomless
-        run: |
-          dotnet new tool-manifest
-          dotnet tool install fantomless-tool --version 4.7.997-prerelease
-          dotnet fantomless --recurse .
-          git diff --exit-code
+      - name: Check style of our F#, TypeScript and YML code
```
knocte, line 147: @realmarv XAML code too, right?

scripts/styleChecker.fsx
realmarv: Yes, that's the case. We never want to raise an error in case of a warning and we want to decide whether to raise an error or not in case of an error.

knocte: then I'd say it's easier to read if you rename the param to `ignoreErrorExitCode`

scripts/styleChecker.fsx
```diff
@@ -0,0 +1,393 @@
...
+
+    if not(isPrettierInstalled) then
+        Process.Execute(
+            {
+                Command = "npm"
+                Arguments = $"install prettier@{version}"
+            },
+            Echo.Off
+        )
+        |> UnwrapProcessResult "" true
```
knocte, line 124: @realmarv whenever you are writing `""`, please please please:
- Either write `String.Empty`, or:
- Use `None` (in this case None is better).

scripts/styleChecker.fsx
```diff
@@ -192,7 +197,14 @@ let StyleXamlFiles() =
     Process
         .Execute(
             {
-                Command = "./node_modules/.bin/prettier"
+                Command =
+                    Path.Combine
+                        [|
+                            "."
```
knocte, line 203: @realmarv wrong, you have to use `Directory.GetCurrentDirectory()` instead of `.`

realmarv: Should I return the used command in StyleXFiles so we can use the command in the suggestion string of CheckStyleOfXFiles?

knocte: no because in the suggestion we want to use `.`

scripts/styleChecker.fsx
realmarv: > @realmarv actually this whole function shouldn't be called if not in CI
why?

.github/workflows/CI.yml

scripts/styleChecker.fsx
realmarv: Done

scripts/styleChecker.fsx
realmarv: Done

scripts/styleChecker.fsx
realmarv: Done

scripts/styleChecker.fsx
realmarv: Done

scripts/styleChecker.fsx
realmarv: If we want to run the script outside CI we first need to make sure there are no uncommitted changes in the local machine

scripts/styleChecker.fsx
realmarv: Done

scripts/styleChecker.fsx
realmarv: Done

realmarv:
> @realmarv also please rebase this PR when you have time (BTW the reason why I thought this PR might become kinda important soon is that yesterday I saw the guys arguing over having too many EOLs in NX repo, like 2 EOLs instead of 1 EOL, which I think it's ridiculous bike-shedding, which can only be fixed by enforcing fantomless, so that those conversations are not needed at all ever again)

Done

scripts/styleChecker.fsx
```diff
@@ -0,0 +1,442 @@
...
+    let prettierPath =
+        Path.Combine
+            [|
+                "."
+                "node_modules"
+                ".bin"
+                "prettier"
+            |]
+
+    let pattern = Path.Combine [| "**"; "*.xaml" |]
```
knocte, line 409: @realmarv a pattern is not a path, please don't use Path.Combine here, use Path.DirectorySeparatorChar

scripts/styleChecker.fsx
knocte: I'm going to make a suggestion for this script and this way you'll understand why: let's split this script in too, one of them should be called styleApply.fsx and the other styleCheck.fsx; the latter should call the former, and only the latter should call git diff (and CI should call the latter)

knocte:
CI still red

scripts/styleChecker.fsx
knocte: why? when a developer runs fantomas or fantomless on his repo, fantomas/fantomless doesn't check git at all; and this .fsx script is just trying to do a similar thing as fantomless, but adjusting style of more files, not just F#

scripts/styleChecker.fsx
```diff
@@ -0,0 +1,442 @@
...
+        .UnwrapDefault()
+    |> ignore
+
+    let pattern = Path.Combine [| "**"; "*.xaml" |]
+
+    Process
+        .Execute(
+            {
+                Command =
+                    Path.Combine
```
knocte, line 203: @realmarv I don't know why you started using Path.Combine as it receives an array, but that's not the way it should be used; Path.Combine receives arguments (C#-style arguments, which, to F# are shown as a tuple), even if the number of arguments is variable; so please just use parentheses when calling Path.Combine

scripts/styleChecker.fsx
realmarv: Done

scripts/styleChecker.fsx
realmarv: Should I add the functions in this script to the FileConventions library and use them in these two scripts? because in the style check part of the script, I'm calling style functions separately to print a proper error message when each case happens

knocte: sounds good, this way there's no need for styleCheck.fsx to call styleApply.fsx

realmarv: not sure if I should add tests for those functions

knocte:
Added some comments in https://github.com/realmarv/conventions/commit/2ecf57bdd5ac6867484a77ddafce73ce9ec956a4

knocte:
And https://github.com/realmarv/conventions/commit/c185305802292b4035a98e645ce2e6df9b8f9b3c

scripts/styleCheck.fsx
```diff
@@ -166,6 +166,19 @@ let CheckStyleOfXamlFiles(rootDir: DirectoryInfo) : bool =
...
+    .Execute(
+        {
+            Command = "chmod"
+            Arguments = "777 --recursive ."
+        },
+        Echo.Off
+    )
+    .UnwrapDefault()
+|> ignore
+
```
knocte, line 181: @realmarv this is not a refactor because you're changing behaviour, you're fixing a bug that I told you, so please squash things properly

knocte:
Added another comment in https://github.com/nblockchain/conventions/pull/103/commits/c7b0b00d054b19ab19d77ff826e113a4a68ec40a

knocte:
> Added some comments in https://github.com/realmarv/conventions/commit/2ecf57bdd5ac6867484a77ddafce73ce9ec956a4

@realmarv please address the comments above. And stop working on another task before you finish this one please. I told you that you shouldn't start new tasks before you finish old ones, and you keep working on a new task that you decided to start some days ago. This PR needs to be merged first before you resume work on that.

BTW how is NX going? I don't see you chasing people to merge PRs and the like. (The fact that I assigned a task to Taras doesn't mean loss of focus on NX, I just gave him one extra low-prio task.)

scripts/styleCheck.fsx
realmarv: I squashed this change to the previous commit

realmarv:
> Added another comment in [c7b0b00](https://github.com/nblockchain/conventions/commit/c7b0b00d054b19ab19d77ff826e113a4a68ec40a)

I don't see any comment there

realmarv:
> > Added some comments in [realmarv@2ecf57b](https://github.com/realmarv/conventions/commit/2ecf57bdd5ac6867484a77ddafce73ce9ec956a4)
> 
> @realmarv please address the comments above. And stop working on another task before you finish this one please. I told you that you shouldn't start new tasks before you finish old ones, and you keep working on a new task that you decided to start some days ago. This PR needs to be merged first before you resume work on that.
> 
Sorry I thought I addressed those comments

knocte:
> I don't see any comment there

Do you see this one?: https://github.com/realmarv/conventions/commit/2ecf57bdd5ac6867484a77ddafce73ce9ec956a4

scripts/styleCheck.fsx
```diff
@@ -0,0 +1,192 @@
...
+            let processResult = GitDiff()
+            UnwrapProcessResult suggestion true processResult |> ignore
+            IsProcessSuccessful processResult
+        else
+            true
+
+    success
+
+let CheckStyleOfXamlFiles(rootDir: DirectoryInfo) : bool =
+    let prettierPath =
```
knocte, line 138: @realmarv this path is not being used by the script, just being shown as an error message to the user; therefore it should not use Directory.GetCurrentDirectory(); it is a relative path that should STAY relative because you want to just show it to the user

knocte:
Before I merge this PR, @aarani can you test the styleApply.fsx script independently?

aarani:
styleApply.fsx doesn't work on windows
```
PS C:\Users\afshi\source\repos\PDNA\PDNA> dotnet fsi C:\conventions\scripts\styleApply.fsx
npm list prettier@2.8.3
Fsdk.Process+ProcessCouldNotStart: Process could not start! Command: npm. Arguments: list prettier@2.8.3.
 ---> System.ComponentModel.Win32Exception (2): An error occurred trying to start process 'npm' with working directory 'C:\Users\afshi\source\repos\PDNA\PDNA'. The system cannot find the file specified.
   at System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)
   at Fsdk.Process.Execute(ProcessDetails procDetails, Echo echo)
   --- End of inner exception stack trace ---
   at Fsdk.Process.Execute(ProcessDetails procDetails, Echo echo)
   at FSI_0003.Helpers.InstallPrettier(String version) in C:\conventions\src\FileConventions\Helpers.fs:line 170
   at FSI_0004.FileConventions.StyleTypeScriptFiles() in C:\conventions\src\FileConventions\Library.fs:line 463
   at <StartupCode$FSI_0005>.$FSI_0005.main@() in C:\conventions\scripts\styleApply.fsx:line 16
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
Stopped due to error
PS C:\Users\afshi\source\repos\PDNA\PDNA> npm --version
9.6.7
PS C:\Users\afshi\source\repos\PDNA\PDNA>
```

aarani:
Wrong commit msg: restore -> resprose

aarani:
Ah also YMI -> YML

knocte:
@parhamsaremi can you test running styleApply.fsx in your Linux OS (which I guess is ubuntu22.04.x) please.

webwarrior-ws:
Superseded by https://github.com/nblockchain/conventions/pull/191

PR #104, opened 2023-04-11
scripts/detectNotUsingGitPush1by1.fsx: fix

realmarv:
In private repositories, this script needs to use an access
token to be able to use GitHub api for accessing repository
info. In this commit I added Authentication header to GitHub
api requests.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -110,6 +110,8 @@ jobs:
     runs-on: ubuntu-22.04
     container:
       image: "ubuntu:22.04"
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
       - name: Install required dependencies
         run: |
--- a/scripts/detectNotUsingGitPush1by1.fsx
+++ b/scripts/detectNotUsingGitPush1by1.fsx
@@ -2,6 +2,7 @@
 
 open System
 open System.IO
+open System.Net
 open System.Net.Http
 open System.Net.Http.Headers
 
@@ -9,7 +10,10 @@ open System.Net.Http.Headers
 open FSharp.Data
 
 #r "nuget: Fsdk, Version=0.6.0--date20230214-0422.git-1ea6f62"
+
 open Fsdk
+open Fsdk.Process
+open Fsdk.FSharpUtil
 
 let githubEventPath = Environment.GetEnvironmentVariable "GITHUB_EVENT_PATH"
 
@@ -18,6 +22,16 @@ if String.IsNullOrEmpty githubEventPath then
         "This script is meant to be used only within a GitHubCI pipeline"
 
     Environment.Exit 2
+
+(*
+Please define GITHUB_TOKEN environment variable in your GitHubCI workflow:
+
+```
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+```
+*)
+let accessToken = Environment.GetEnvironmentVariable "GITHUB_TOKEN"
 
 type githubEventType =
     JsonProvider<"""
@@ -779,7 +793,24 @@ type PRCommitsType =
   ]
 """>
 
-let GitHubApiCall(url: string) =
+let githubApiCallForbiddenErrorMsg =
+    """GITHUB_TOKEN passed doesn't seem to have enough permissions.
+To modify the permissions of your token, navigate to the Settings section of your 
+repository or organization and click on Actions button, then select General. From 
+'Workflow permissions' section on that page, choose 'Read and write permissions' 
+(which grants access to content and the ability to make changes).
+"""
+
+let githubApiCallNotFoundErrorMsg =
+    """Please define GITHUB_TOKEN environment variable in your GitHubCI workflow:
+
+```
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+```
+"""
+
+let GitHubApiCall (url: string) (accessToken: string) =
     let userAgent = ".NET App"
     let xGitHubApiVersion = "2022-11-28"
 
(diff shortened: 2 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -99,6 +99,8 @@ jobs:
     runs-on: ubuntu-22.04
     container:
       image: "ubuntu:22.04"
+    env:
+      ACCESS_TOKEN: ${{ secrets.ACCESS_TOKEN }}
```
parhamsaremi, line 103: Couldn't we use secrets in the code directly? If not, is this safe? I think it should be just didn't know if this is a normal thing to do.

scripts/detectNotUsingGitPush1by1.fsx
```diff
@@ -19,6 +19,37 @@ if String.IsNullOrEmpty githubEventPath then
...
+let accessToken = Environment.GetEnvironmentVariable "ACCESS_TOKEN"
+(*
+To save your GitHub access token as an environment variable in GitHub Actions, you can use the secrets feature in GitHub.
+
+Here are the steps to save your access token as an environment variable:
+
+1. Generate a personal access token on GitHub (if you haven't already done so).
+2. Go to the settings for your repository and click on "Secrets and variables" in the left-hand sidebar.
+3. Click on the "Actions" under "Secrets and variables".
+4. Click on the "New secret" button.
```
parhamsaremi, line 31: New secret -> New repository secret

scripts/detectNotUsingGitPush1by1.fsx
```diff
@@ -19,6 +19,37 @@ if String.IsNullOrEmpty githubEventPath then
...
+5. Enter "ACCESS_TOKEN" as a name for your secret and paste your access token into the "Value" field.
```
parhamsaremi, line 32: into the "Secret" field.

.github/workflows/CI.yml
realmarv: I tried `Environment.GetEnvironmentVariable "secrets.GITHUB_TOKEN"` and `Environment.GetEnvironmentVariable "GITHUB_TOKEN"` but they didn't work. I didn't find a way to do this. Do you know a way to use secrets in the code directly without defining it as an environment variable?

scripts/detectNotUsingGitPush1by1.fsx
realmarv: With secrets.GITHUB_TOKEN we don't need those lines anymore

.github/workflows/CI.yml
parhamsaremi: No, I don't know any way for this. But thinking more about this I think the way you've implemented this should be ok. Because I'm not sure if accessing the secrets from a script is logical anyway. :+1:

scripts/detectNotUsingGitPush1by1.fsx
```diff
@@ -19,6 +19,29 @@ if String.IsNullOrEmpty githubEventPath then
...
+name: My Workflow
+
+on: [push, pull_request]
+
+jobs:
+  my_job:
+    runs-on: ubuntu-latest
+    env:
+      ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
```
parhamsaremi, line 39: Is this part needed to be in the example? Apart from this, it LGTM.

.github/workflows/CI.yml
knocte: where do we stand here?

scripts/detectNotUsingGitPush1by1.fsx
realmarv: Removed it

.github/workflows/CI.yml
realmarv: @knocte Do you think defining an environment variable is the right way to use GITHUB_TOKEN in the script?

realmarv: Yes

knocte: Why don't you share?

realmarv: We can pass the access token as an input argument to the script but I don't think it's a better way than this one

knocte: let's leave it like this then, so where do we stand with this PR? is it tested?

realmarv: Yes it's tested

knocte: And when testing this, you didn't need to add any secret right? it will just work out of the box?

realmarv: Right. It should work

scripts/detectNotUsingGitPush1by1.fsx
```diff
@@ -19,6 +19,26 @@ if String.IsNullOrEmpty githubEventPath then
 
     Environment.Exit 2
 
+let accessToken = Environment.GetEnvironmentVariable "ACCESS_TOKEN"
```
knocte, line 22: @realmarv shouldn't this line be below the comment instead of above it?

knocte:
@realmarv actually there's something else that should be done in this PR: detect the 404 Not Found error in case it happens, and prompt with an error message to suggest to add this to GitHubCI:
```
    env:
      ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

scripts/detectNotUsingGitPush1by1.fsx
```diff
@@ -794,11 +815,47 @@ let GitHubApiCall(url: string) =
...
+        )
+
+    let result =
+        try
+            (client.GetStringAsync url).Result
+        with
+        | :? AggregateException as aggregateException ->
+            match aggregateException.InnerException with
+            | :? HttpRequestException as httpRequestException ->
+                match httpRequestException.StatusCode with
```
knocte, line 831: @realmarv this is a nullable object, an ugly C# equivalent of `Option`; the best way to deal with it from F# is pass it to Option.ofObj, so `match httpRequestException.StatusCode |> Option.ofObj`, then you can use `| Some statusCode ->` so that you don't need to call `HasValue` and `Value` at all

scripts/detectNotUsingGitPush1by1.fsx
```diff
@@ -794,11 +815,47 @@ let GitHubApiCall(url: string) =
...
+                | statusCode ->
+                    if statusCode.HasValue
+                       && statusCode.Value = HttpStatusCode.NotFound then
+                        let errorMsg =
+                            """
+Please add the \"ACCESS_TOKEN\" environment variable to the GitHubCI file:
```
knocte, line 837: @realmarv this text has to start after the 3 double-quote chars, not after an EOL

knocte (changes requested):

scripts/detectNotUsingGitPush1by1.fsx
realmarv: With this modification I get this error: `The type 'Nullable<HttpStatusCode>' does not have 'null' as a proper value. To create a null value for a Nullable type use 'System.Nullable()'.` and I don't know what to do with it

knocte: looks good, it's what I meant

realmarv: But I get the error I mentioned

scripts/detectNotUsingGitPush1by1.fsx

knocte:
@realmarv what's going on here? why is this not finished and why is there a WIP1 commit in the PR and why CI is red? Actually this has more priority than typescript work because it is previous unfinished work.

.github/workflows/CI.yml
```diff
@@ -99,6 +99,8 @@ jobs:
     runs-on: ubuntu-22.04
     container:
       image: "ubuntu:22.04"
+    env:
+      ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
knocte, line 103: @realmarv haven't we agreed in other PRs to call this env GITHUB_TOKEN instead of ACCESS_TOKEN? let's fix it here too

realmarv:
> @realmarv what's going on here? why is this not finished and why is there a WIP1 commit in the PR and why CI is red? Actually this has more priority than typescript work because it is previous unfinished work.

Done

knocte:
@realmarv unfortunately CI in your fork is red, and I believe it's because of a recent issue we had wrt commitlint, which should be fixed if you rebase your branch.

realmarv:
> @realmarv unfortunately CI in your fork is red, and I believe it's because of a recent issue we had wrt commitlint, which should be fixed if you rebase your branch.

Done

PR #105, opened 2023-04-11
Add a script to wrap the last commit msg body into a certain character count per line

realmarv:
Added the wrapLatestCommitMsg.fsx script to wrap the last commit
message while considering code blocks and footer notes including
references, Fixes and Closes and Co-authored by lines.

### DIFF
--- a/commitlint/plugins.ts
+++ b/commitlint/plugins.ts
@@ -403,7 +403,8 @@ export abstract class Plugins {
 
     public static bodyParagraphLineMinLength(
         bodyStr: string | null,
-        paragraphLineMinLength: number
+        paragraphLineMinLength: number,
+        paragraphLineMaxLength: number
     ) {
         let offence = false;
 
@@ -414,14 +415,14 @@ export abstract class Plugins {
             for (let paragraph of paragraphs) {
                 let lines = paragraph.split(/\r?\n/);
                 let inBigBlock = false;
-                for (let i = 0; i < lines.length; i++) {
+                for (let i = 0; i < lines.length - 1; i++) {
                     let line = lines[i];
 
                     if (Helpers.isBigBlock(line)) {
                         inBigBlock = !inBigBlock;
                         continue;
                     }
-                    if (inBigBlock || i === lines.length - 1) {
+                    if (inBigBlock) {
                         continue;
                     }
 
@@ -434,11 +435,12 @@ export abstract class Plugins {
 
                         let lineIsFooterNote = Helpers.isFooterNote(line);
 
-                        if (
-                            !isUrl &&
-                            !lineIsFooterNote &&
-                            line !== lines[lines.length - 1]
-                        ) {
+                        let nextWordLength = lines[i + 1].split(" ")[0].length;
+                        let isNextWordTooLong =
+                            nextWordLength + line.length + 1 >
+                            paragraphLineMaxLength;
+
+                        if (!isUrl && !lineIsFooterNote && !isNextWordTooLong) {
                             offence = true;
                             break;
                         }
--- a/commitlint/plugins.ts
+++ b/commitlint/plugins.ts
@@ -401,6 +401,60 @@ export abstract class Plugins {
         ];
     }
 
+    public static bodyParagraphLineMinLength(
+        bodyStr: string | null,
+        paragraphLineMinLength: number
+    ) {
+        let offence = false;
+
+        if (bodyStr !== null) {
+            bodyStr = Helpers.removeAllCodeBlocks(bodyStr).trim();
+
+            let paragraphs = bodyStr.split(/\r?\n\r?\n/);
+            for (let paragraph of paragraphs) {
+                let lines = paragraph.split(/\r?\n/);
+                let inBigBlock = false;
+                for (let i = 0; i < lines.length; i++) {
+                    let line = lines[i];
+
+                    if (Helpers.isBigBlock(line)) {
+                        inBigBlock = !inBigBlock;
+                        continue;
+                    }
+                    if (inBigBlock || i === lines.length - 1) {
+                        continue;
+                    }
+
+                    let nextLine = lines[i + 1];
+
+                    if (line.length < paragraphLineMinLength) {
+                        let isUrl =
+                            Helpers.isValidUrl(line) ||
+                            Helpers.isValidUrl(nextLine);
+
+                        let lineIsFooterNote = Helpers.isFooterNote(line);
+
+                        if (
+                            !isUrl &&
+                            !lineIsFooterNote &&
+                            line !== lines[lines.length - 1]
+                        ) {
+                            offence = true;
+                            break;
+                        }
+                    }
+                }
+            }
+        }
+
+        return [
+            !offence,
+            `Please do not subceed ${paragraphLineMinLength} characters in the lines of the commit message's body paragraphs (except the last line of each paragraph); we recommend this script (for editing the last commit message): \n` +
+                "https://github.com/nblockchain/conventions/blob/master/scripts/wrapLatestCommitMsg.fsx" +
+                Helpers.errMessageSuffix,
+        ];
+    }
+
     public static trailingWhitespace(rawStr: string) {
         let offence = false;
 
--- a/commitlint/plugins.ts
+++ b/commitlint/plugins.ts
@@ -393,16 +393,10 @@ export abstract class Plugins {
             }
         }
 
-        // taken from https://stackoverflow.com/a/66433444/544947 and https://unix.stackexchange.com/a/25208/56844
-        function getUnixCommand(fmtOption: string) {
-            return `git log --format=%B -n 1 $(git log -1 --pretty=format:"%h") | cat - > log.txt ; fmt -w 1111 -s log.txt > ulog.txt && fmt -w 64 -s ${fmtOption} ulog.txt > wlog.txt && git commit --amend -F wlog.txt`;
-        }
-
         return [
             !offence,
-            `Please do not exceed ${bodyMaxLineLength} characters in the lines of the commit message's body; we recommend this unix command (for editing the last commit message): \n` +
-                `For Linux users: ${getUnixCommand("-u")}\n` +
-                `For macOS users: ${getUnixCommand("")}` +
+            `Please do not exceed ${bodyMaxLineLength} characters in the lines of the commit message's body; we recommend this script (for editing the last commit message): \n` +
+                "https://github.com/nblockchain/conventions/blob/master/scripts/wrapLatestCommitMsg.fsx" +
                 Helpers.errMessageSuffix,
         ];
     }
--- a/scripts/wrapLatestCommitMsg.fsx
+++ b/scripts/wrapLatestCommitMsg.fsx
@@ -1,0 +1,67 @@
+#!/usr/bin/env -S dotnet fsi
+
+open System.IO
+open System
+open System.Text.RegularExpressions
+
+#load "../src/FileConventions/Library.fs"
+
+#r "nuget: Fsdk, Version=0.6.0--date20230214-0422.git-1ea6f62"
+
+open Fsdk
+open Fsdk.Process
+
+let commitMsg =
+    Fsdk
+        .Process
+        .Execute(
+            {
+                Command = "git"
+                Arguments = "log -1 --format=%B"
+            },
+            Echo.Off
+        )
+        .UnwrapDefault()
+        .Trim()
+
+let header, maybeBody =
+    let newLineIndex = commitMsg.IndexOf Environment.NewLine
+
+    if newLineIndex > 0 then
+        commitMsg.Substring(0, newLineIndex).Trim(),
+        Some(commitMsg.Substring(newLineIndex).Trim())
+    else
+        commitMsg, None
+
+let maxCharsPerLine = 64
+
+let maybeWrappedBody =
+    match maybeBody with
+    | Some body -> Some(FileConventions.WrapText body maxCharsPerLine)
+    | _ -> None
+
+let EscapeDoubleQuotes(text: string) =
+    Regex.Replace(text, @"([^\\])""", @"$1\""")
+
+let newCommitMsg =
+    match maybeWrappedBody with
+    | Some wrappedBody ->
+        header
+        + Environment.NewLine
+        + Environment.NewLine
+        + wrappedBody
+        + Environment.NewLine
+    | _ -> header
+
+Fsdk
+    .Process
+    .Execute(
+        {
+            Command = "git"
+            Arguments =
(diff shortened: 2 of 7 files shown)


scripts/detectNotUsingGitPush1by1.fsx
```diff
@@ -536,53 +536,283 @@ let gitForkUser = parsedJsonObj.PullRequest.Head.User.Login
...
-        client.DefaultRequestHeaders.Add("X-GitHub-Api-Version", "2022-11-28")
 
         let url =
             sprintf
                 "https://api.github.com/repos/%s/commits/%s/check-suites"
                 gitRepo
                 commit
 
-        let json = (client.GetStringAsync url).Result
+        let json = GitHubApiCall url
```
knocte, line 815: @realmarv changes in this file look unrelated to this PR

realmarv:
> Please change the code in our commitlint plugins where we advice to use a linux/mac command.

Done

src/FileConventions/Library.fs
```diff
@@ -104,3 +104,100 @@ let EolAtEof(fileInfo: FileInfo) =
...
+                        line.Split(" ")
+                        |> Seq.map(fun word -> Text(word, PlainText))
+                )
+                |> Seq.concat
+        )
+        |> Seq.concat
+
+    words |> Seq.toList
+
+let WrapParagraph (paragraph: string) (maxCharsPerLine: int) : string =
```
knocte, line 169: @realmarv this func should be private

src/FileConventions/Library.fs
```diff
@@ -104,3 +104,100 @@ let EolAtEof(fileInfo: FileInfo) =
...
+    IsFooterReference line
+    || IsCoAuthoredByTag line
+    || IsFixesOrClosesSentence line
+
+type Word =
+    | CodeBlock
+    | FooterNote
+    | PlainText
+
+type Text(textStr: string, textType: Word) =
```
knocte, line 127: @realmarv there's no reason for why this type cannot be a record instead of a class (records are preferred in F#, classes should be seldomly used, because records are idiomatic F#), unless I'm missing something

src/FileConventions/Library.fs
realmarv: How should I test a private function?

knocte: you shouldn't test WrapParagraph, you should just test WrapText

src/FileConventions/Library.fs

src/FileConventions.Test/FileConventions.Test.fs
```diff
@@ -255,3 +255,78 @@ let HasBinaryContentTest3() =
...
+    )
+
+[<Test>]
+let WrapParagraphTest3() =
+    let characterCount = 64
+
+    let paragraph =
+        "This is short line referring to [1]."
+        + Environment.NewLine
+        + "[1] This is a very very very very long line with more than 64 characters"
```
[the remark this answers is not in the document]
realmarv: Done

src/FileConventions/Library.fs
```diff
@@ -104,3 +104,112 @@ let EolAtEof(fileInfo: FileInfo) =
...
+        )
+        |> Seq.concat
+
+    words |> Seq.toList
+
+let private WrapParagraph (text: string) (maxCharsPerLine: int) : string =
+    let words = SplitIntoWords text
+
+    let mutable currentLine = ""
+    let mutable wrappedText = ""
```
knocte, line 185: @realmarv please rewrite the algorithm with no mutable vars, this is not idiomatic F#

src/FileConventions/Library.fs
```diff
@@ -104,3 +104,112 @@ let EolAtEof(fileInfo: FileInfo) =
...
+
+    for word in words do
+        if String.IsNullOrEmpty(currentLine) then
```
knocte, line 188: @realmarv nit: no need for parens

src/FileConventions/Library.fs
```diff
@@ -104,3 +104,112 @@ let EolAtEof(fileInfo: FileInfo) =
...
+            currentLine <- currentLine + " " + word.Text
+        else
+            wrappedText <- wrappedText + currentLine + Environment.NewLine
+            currentLine <- word.Text
+
+    (wrappedText + currentLine).Trim()
+
+let WrapText (text: string) (maxCharsPerLine: int) : string =
+    let wrappedParagraphs =
+        text.Split($"{Environment.NewLine}{Environment.NewLine}")
```
knocte, line 209: @realmarv nit: no need for parens

src/FileConventions/Library.fs
```diff
@@ -104,3 +104,112 @@ let EolAtEof(fileInfo: FileInfo) =
...
+                |> Seq.map(fun line ->
+                    if IsFooterNote line then
+                        Seq.singleton(
+                            {
+                                Text = line
+                                Type = FooterNote
+                            }
+                        )
+                    else
+                        line.Split(" ")
```
knocte, line 167: @realmarv nit: no need for parens

src/FileConventions/Library.fs

src/FileConventions/Library.fs

src/FileConventions/Library.fs
realmarv: If you like the new function I can squash it

knocte: no need to squash, it can count as a refactoring commit

src/FileConventions/Library.fs
```diff
@@ -181,28 +181,43 @@ let SplitIntoWords(text: string) =
...
+              {
+                  Type = PlainText
+              } ->
+                processWords
+                    word.Text
+                    (wrappedText + currentLine + Environment.NewLine)
+                    rest
+            | _, _ ->
+                processWords
+                    ""
```
knocte, line 212: @realmarv please use String.Empty

realmarv:
> This PR should also contain the implementation of a commitlint rule that detects when lines in the body are too short (less than headerMaxLineLength which currently is 50) and suggest this script (in the same way that the other rule bodySoftMaxLineLength does). This plugin should of course be developed using TDD as well, and ignore the last line of a paragraph. Maybe let's call it bodyParagraphLineMinLength?

Done

commitlint/plugins.ts
```diff
@@ -401,6 +401,60 @@ export abstract class Plugins {
...
+                            break;
+                        }
+                    }
+                }
+            }
+        }
+
+        return [
+            !offence,
+            `Please do not subceed ${paragraphLineMinLength} characters in the lines of the commit message's body paragraphs (Except the last line of each paragraph); we recommend this script (for editing the last commit message): \n` +
```
knocte, line 452: @realmarv why did you make the word Except start with uppercase letter?

realmarv: fixed it

knocte:
@realmarv we need a new testcase for this rule:

```
Some title of less than 50 chars

This is a paragraph whose 2nd line is less than 50 chars
but should not make commitlint complain because
TheNextWordInThe3rdLineIsTooLongToBePlacedIn2ndLine ok? 
```

Do you understand?

realmarv:
> @realmarv we need a new testcase for this rule:
> 
> ```
> Some title of less than 50 chars
> 
> This is a paragraph whose 2nd line is less than 50 chars
> but should not make commitlint complain because
> TheNextWordInThe3rdLineIsTooLongToBePlacedIn2ndLine ok? 
> ```
> 
> Do you understand?

That makes sense

realmarv:
> @realmarv we need a new testcase for this rule:
> 
> ```
> Some title of less than 50 chars
> 
> This is a paragraph whose 2nd line is less than 50 chars
> but should not make commitlint complain because
> TheNextWordInThe3rdLineIsTooLongToBePlacedIn2ndLine ok? 
> ```
> 
> Do you understand?

Done, but I had to remove the input variable to the rule and use the global variables because using two input variables (header max length and body max length) causes failure of all tests which probably means the code doesn't compile and I think it's probably a commitlint issue which means I need to investigate the commitlint repo if you agree.

knocte:
>  I think it's probably a commitlint issue which means I need to investigate the commitlint repo if you agree.

I don't think it's worth looking into this.

scripts/wrapLatestCommitMsg.fsx
```diff
@@ -0,0 +1,46 @@
...
+        .Trim()
+
+let header, body =
+    let newLineIndex = commitMsg.IndexOf(Environment.NewLine)
+
+    if newLineIndex > 0 then
+        commitMsg.Substring(0, newLineIndex).Trim(),
+        commitMsg.Substring(newLineIndex).Trim()
+    else
+        commitMsg, null
```
knocte, line 33: @realmarv null is non-idiomatic F#, please only use null when you need to call C#

scripts/wrapLatestCommitMsg.fsx
```diff
@@ -0,0 +1,46 @@
...
+                Command = "git"
+                Arguments = "log -1 --format=%B"
+            },
+            Echo.Off
+        )
+        .UnwrapDefault()
+        .Trim()
+
+let header, body =
+    let newLineIndex = commitMsg.IndexOf(Environment.NewLine)
```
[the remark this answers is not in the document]
realmarv: Done

scripts/wrapLatestCommitMsg.fsx
realmarv: Done

scripts/wrapLatestCommitMsg.fsx
```diff
@@ -0,0 +1,49 @@
...
+        Some(commitMsg.Substring(newLineIndex).Trim())
+    else
+        commitMsg, None
+
+let maxCharsPerLine = 64
+
+let wrappedBody =
+    match mayBeBody with
+    | Some body -> FileConventions.WrapText body maxCharsPerLine
+    | _ -> ""
```
knocte, line 40: @realmarv please use String.Empty, not `""`

scripts/wrapLatestCommitMsg.fsx
```diff
@@ -0,0 +1,49 @@
...
+    if newLineIndex > 0 then
+        commitMsg.Substring(0, newLineIndex).Trim(),
+        Some(commitMsg.Substring(newLineIndex).Trim())
+    else
+        commitMsg, None
+
+let maxCharsPerLine = 64
+
+let wrappedBody =
+    match mayBeBody with
```
knocte, line 38: @realmarv nit: should be `maybeBody`

scripts/wrapLatestCommitMsg.fsx
```diff
@@ -0,0 +1,49 @@
...
+    | _ -> ""
+
+let newCommitMsg =
+    header
+    + Environment.NewLine
+    + Environment.NewLine
+    + wrappedBody
+    + Environment.NewLine
+
+File.WriteAllText("wlog.txt", newCommitMsg)
```
knocte, line 49: @realmarv this is wrong, because the previous message we gave to the user involved dealing with this wlog.txt, but then now you tell the user to call this script and you don't tell him what to do with this file

scripts/wrapLatestCommitMsg.fsx
realmarv: Done

scripts/wrapLatestCommitMsg.fsx
realmarv: Done

scripts/wrapLatestCommitMsg.fsx
realmarv: I added the git amend command. BTW I'll squash when we're done with this script

realmarv: What about not creating a file at all?

knocte: that would be even better indeed; if you find a way to do that, great

realmarv: git amend has a -m option for defining the commit message directly

knocte: Great let's use it

realmarv: Done

knocte: tested?

scripts/wrapLatestCommitMsg.fsx
```diff
@@ -0,0 +1,59 @@
...
+    + Environment.NewLine
+    + wrappedBody
+    + Environment.NewLine
+
+Fsdk
+    .Process
+    .Execute(
+        {
+            Command = "git"
+            Arguments = $"commit --amend -m \"{newCommitMsg}\""
```
knocte, line 54: @realmarv btw should use a flag with a word, not with a single letter; and I asked if you tested it because I wonder what would happen with commit msgs that contain double quotes?

realmarv: triple quotes can be used for defining a raw string in git command but it's not working with Fsdk for some reason

realmarv: maybe it's just easier to get back to the temporary file option

knocte: triple quotes is an F# syntax, you're dealing here with commandline related things, something entirely different; isn't it just enough to replace all double-quote chars with an escaped double-quote?

realmarv: Done

knocte:
So this is properly squashed and tested right? If yes, I'll merge.

knocte:
Oh, last thing is deleting a TODO line about detecting paragraphs that are cropped too short.

realmarv:
> So this is properly squashed and tested right? If yes, I'll merge.

Actually, I noticed that the script is removing double quotes and the [^\\] pattern is not working in F# regex. This is the pattern `"[^\\]\""` I'm using for finding double quotes that are not following a backslash character

knocte:
> Actually, I noticed that the script is removing double quotes and the [^\\] pattern is not working in F# regex. This is the pattern "[^\\]\"" I'm using for finding double quotes that are not following a backslash character

I'm not sure I understand this answer. And anyway, this doesn't seem to answer my question directly. Answer my question directly first, and later you can add more details like the above. BTW CI is red.

realmarv:
> > Actually, I noticed that the script is removing double quotes and the [^\\] pattern is not working in F# regex. This is the pattern "[^\\]"" I'm using for finding double quotes that are not following a backslash character
> 
> I'm not sure I understand this answer. And anyway, this doesn't seem to answer my question directly. Answer my question directly first, and later you can add more details like the above. BTW CI is red.

It's squashed and tested now

realmarv:
> Oh, last thing is deleting a TODO line about detecting paragraphs that are cropped too short.

Done

PR #111, opened 2023-05-31
Fix error message for footernote references with EOL

parhamsaremi:
When we have a commit with EOL after "[]" then we get a confusing error about the reference not existing. In this PR I first check if a reference number has no reference and give a different error about it.

### DIFF
--- a/commitlint/helpers.ts
+++ b/commitlint/helpers.ts
@@ -98,6 +98,15 @@ export abstract class Helpers {
         return isLowerCase && !isUpperCase;
     }
 
+    public static isEmptyFooterReference(line: string) {
+        Helpers.assertLine(line);
+        let trimmedLine = line.trim();
+        return (
+            trimmedLine[0] === "[" &&
+            trimmedLine.indexOf("]") === trimmedLine.length - 1
+        );
+    }
+
     public static isFooterReference(line: string) {
         Helpers.assertLine(line);
         return line[0] === "[" && line.indexOf("] ") > 0;
--- a/commitlint/plugins.test.ts
+++ b/commitlint/plugins.test.ts
@@ -476,6 +476,19 @@ test("footer-references-existence4", () => {
     expect(footerReferenceExistence4.status).toBe(0);
 });
 
+test("footer-references-with-EOL", () => {
+    let commmitMsgwithEOLFooter =
+        "foo: this is only a title" +
+        "\n\n" +
+        "Bla bla blah[1]." +
+        "\n\n" +
+        "[1]\nhttp://foo.bar/baz";
+    let footerReferenceWithEOL = runCommitLintOnMsg(commmitMsgwithEOLFooter);
+    expect(footerReferenceWithEOL.output[1].toString().includes("EOL")).toBe(
+        true
+    );
+});
+
 test("prefer-slash-over-backslash1", () => {
     let commitMsgWithBackslash = "foo\\bar: bla bla bla";
     let preferSlashOverBackslash1 = runCommitLintOnMsg(commitMsgWithBackslash);
--- a/commitlint.config.ts
+++ b/commitlint.config.ts
@@ -114,7 +114,7 @@ module.exports = {
                 "footer-references-existence": ({ body }: { body: any }) => {
                     let bodyStr = Helpers.convertAnyToString(body, "body");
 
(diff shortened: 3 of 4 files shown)


commitlint/plugins.test.ts
```diff
@@ -476,6 +476,17 @@ test("footer-references-existence4", () => {
     expect(footerReferenceExistence4.status).toBe(0);
 });
 
+test("footer-references-with-EOL", () => {
+    let commmitMsgwithEOLFooter =
+        "foo: this is only a title" +
+        "\n\n" +
+        "Bla bla blah[1]" +
```
knocte, line 483: @parhamsaremi this commit msg will make the paragraph rule fail

commitlint/helpers.ts
```diff
@@ -98,6 +98,11 @@ export abstract class Helpers {
         return isLowerCase && !isUpperCase;
     }
 
+    public static isEmptyFooterReference(line: string) {
+        Helpers.assertLine(line);
+        return line[0] === "[" && line.indexOf("]") > 0 && line.trim().length === 3;
```
knocte, line 103: @parhamsaremi hardcoding 3 here is still wrong, not sure if you understood my previous comment

parhamsaremi: hmmm right. I understand now. I'll check with regex to see if it's only footer ref number.

knocte: I don't think using regex here is worth it, just check if the string last char before EOL (trimmed) is `]`

parhamsaremi: :+1:

commitlint/plugins.test.ts
parhamsaremi: I added a . after Bla bla.

commitlint/helpers.ts
parhamsaremi: In the end I just check if the indexof "]" is length-1

parhamsaremi: I think I have to use regex to catch this :thinking:  do you agree? @knocte

parhamsaremi: I changed it to regex and also there was some sanity check errors that I addressed. let me know if you still think I shouldn't use regex and checking last index to be "]" is ok.

knocte: > Although now that I think about it what if the user has something like this?

Such an edge case doesn't deserve the complexity of the code that a regex brings. If we find an edge case like that we will develop a PR on its own with its own unit test.

knocte: There are thousands or millions of bugs lurking in any piece of software, you cannot expect all of them in just one line of code. We work on a 1 by 1 basis.

parhamsaremi: makes sense. I'll bring it back

knocte:
@aarani please review

commitlint/plugins.test.ts
```diff
@@ -476,6 +476,19 @@ test("footer-references-existence4", () => {
...
 
+test("footer-references-with-EOL", () => {
+    let commmitMsgwithEOLFooter =
+        "foo: this is only a title" +
+        "\n\n" +
+        "Bla bla blah[1]." +
+        "\n\n" +
+        "[1]\nhttp://foo.bar/baz";
+    let footerReferenceWithEOL = runCommitLintOnMsg(commmitMsgwithEOLFooter);
+    expect(footerReferenceWithEOL.output[1].toString().includes("EOL")).toBe(
```
aarani, line 487: This expectation looks kinda weird

aarani: Isn't a better way to "expect" this?

aarani: What if we change the message later?

parhamsaremi: That was the best way I could think of. We are sure that there is an error we just want to check if the message is what we expect. Do you have any suggestions?

aarani: > That was the best way I could think of. We are sure that there is an error we just want to check if the message is what we expect. Do you have any suggestions?

true, error always happens here so not really, I don't have a suggestion, I guess this is our only option then.

parhamsaremi: I agree that it's not ideal and EOL was the shortest thing we could think of that may never change in the error message :sweat_smile:

Issue #123, opened 2023-07-12
wrapLatestCommitMsg.fsx doesn't seem to work in Windows


knocte:
My latest commit msg is:

```
GitHubCI: move release creation outta linux job

Given that the asset upload fails often, and needs to be retried;
then it's better to put release creation out of the job that
needs to be retried (otherwise it fails with "release already
exists", and then the release needed to be deleted manually
before the retry).
```

First line is over 64 chars and I call wrapLatestCommit.fsx and nothing changes.

PR #126, opened 2023-07-13
wrapLatestCommitMsg: fix behaviour in windows

aarani:
While putting Environment.NewLine in output doesn't cause any problems expecting Environment.NewLine in input commit msgs causes problem in windows, so while keeping Environment.NewLine for output we change input processing to split by unix newline.

Fixes https://github.com/nblockchain/conventions/issues/123

### DIFF
--- a/src/FileConventions/Library.fs
+++ b/src/FileConventions/Library.fs
@@ -5,6 +5,35 @@ open System.IO
 open System.Linq
 open System.Text.RegularExpressions
 
+// While putting Environment.NewLine in output doesn't cause any problems
+// expecting Environment.NewLine in input commit msgs causes problem in
+// windows, so while keeping Environment.NewLine for output we change
+// input processing to split by \n (Unix NewLine)
+let UnixEol = "\n"
+let WindowsEol = "\r\n"
+let MacEol = "\r"
+
+let SplitByEOLs
+    (text: string)
+    (numberOfEOLs: int)
+    (splitOptions: StringSplitOptions)
+    (acceptablePartCount: Option<int>)
+    =
+    if numberOfEOLs < 0 then
+        failwith "NumberOfEOLs should be more than zero"
+
+    let acceptablePartCount =
+        Option.defaultValue Int32.MaxValue acceptablePartCount
+
+    let preparedText =
+        text
+            .Replace(WindowsEol, UnixEol)
+            .Replace(MacEol, UnixEol)
+
+    let separator = String.replicate numberOfEOLs UnixEol
+
+    preparedText.Split(separator, acceptablePartCount, splitOptions)
+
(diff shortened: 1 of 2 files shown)


knocte:
@realmarv please review

src/FileConventions/Library.fs
```diff
@@ -238,7 +249,13 @@ let private WrapParagraph (text: string) (maxCharsPerLine: int) : string =
 
 let WrapText (text: string) (maxCharsPerLine: int) : string =
     let wrappedParagraphs =
-        text.Split $"{Environment.NewLine}{Environment.NewLine}"
+        text.Split(
+            [|
+                $"{UnixEol}{UnixEol}"
+                $"{Environment.NewLine}{Environment.NewLine}"
```
realmarv, line 255: why using UnixEol and Environment.NewLine instead of UnixEol and WindowsEol?

aarani: I made it more consistant

knocte: @aarani now that we are making changes to this, let's also make sure it works in macOS, and to do that I'd prefer if we adopt a common function instead of violating DRY, let's do something similar to this: https://github.com/nblockchain/conventions/commit/2ccc084d8187913decfac4b7f893010612b5bfce

src/FileConventions/Library.fs
```diff
@@ -169,7 +198,14 @@ let SplitIntoWords(text: string) =
...
                 Seq.singleton paragraph
             else
-                let lines = paragraph.Text.Split Environment.NewLine
+                let separatorEolCount = 1
+
+                let lines =
+                    SplitByEOLs
+                        paragraph.Text
+                        separatorEolCount
+                        StringSplitOptions.None
```
knocte, line 207: @aarani are you sure RemoveEmptyEntries doesnt work here???

aarani: no I'm not sure.

aarani: RemoveEmptyEntries by nature changes the input which I don't think is the intended behaviour here. Only place I want to use RemoveEmptyEntires is to separate header and body from the commit msg

knocte: it changes the input?? sorry, what???

knocte: @aarani ping

aarani: > it changes the input?? sorry, what???

what RemoveEmptyEntries does is that it removes empty strings in the split result, since we are joining them back together removing empty entires changes the output result which we don't want here.

knocte:
Sorry Afshin, this PR didn't respect the KISS principle ;) I submitted a simpler fix here: 4c7790fd217c92e8f51e28aef84d056a29094674

Issue #118, opened 2023-07-06
footer-notes-misplacement false positive


knocte:
From https://github.com/nodeeffect/RunIntoMe/actions/runs/5471278315/jobs/9962255316:

````
npx commitlint --from HEAD~1 --to HEAD --verbose
⧗   input: GitHubCI: rollback prettier/plugin-xml to 2.2.0

We had a CI regression [1] when formatting XAML files:

```
sudo npm install --save-dev prettier @prettier/plugin-xml
added 5 packages in 898ms
[warn] Ignored unknown option --xml-whitespace-sensitivity=ignore. Did you mean --html-whitespace-sensitivity?
Error:  No parser could be inferred for file "/__w/RunIntoMe/RunIntoMe/src/Frontend/AddFriendPage.xaml".
...
```

[1] https://github.com/nodeeffect/RunIntoMe/actions/runs/5466042118/jobs/9959987069
✖   Footer messages must be placed after body paragraphs, please move any message that starts with "Fixes", "Closes" or "[i]" to the end of the commmit message.
For reference, these are the guidelines that include our commit message conventions: https://github.com/nblockchain/conventions/blob/master/WorkflowGuidelines.md [footer-notes-misplacement]
````

PR #129, opened 2023-07-24
Footer note misplacement fix false positive

parhamsaremi:
Fixes https://github.com/nblockchain/conventions/issues/118

### DIFF
--- a/commitlint/plugins.test.ts
+++ b/commitlint/plugins.test.ts
@@ -529,6 +529,17 @@ test("footer-notes-misplacement-4", () => {
     expect(footerNotesMisplacement4.status).not.toBe(0);
 });
 
+test("footer-notes-misplacement-5", () => {
+    let commitMsgWithRightFooter =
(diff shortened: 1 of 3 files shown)


commitlint/plugins.test.ts
```diff
@@ -534,9 +534,9 @@ test("footer-notes-misplacement-5", () => {
         "foo: this is only a title\n\n" +
         "Bla bla blah[1]:\n\n" +
         "```\nSome error message with a [] in the first of its line\n[warn] some warning\n```\n\n" +
-        "[1] http://foo.bar/baz"
+        "[1] http://foo.bar/baz";
```
parhamsaremi, line 537: These are in the wrong commit, I should fix these.

knocte:
> Fix false positive mentioned in the https://github.com/nblockchain/conventions/issues/118.

Why is this not written in the commit msg of the last commit? /cc @webwarrior-ws

Issue #124, opened 2023-07-12
body-soft-max-line-length exception


knocte:
This rule should not break the build if the offending line exceeds 64 chars because it uses a commit hash at the end.

Testcase:
```
GrpcService: fix some logging nits

These mistakes were made in 45faeca2f0e7c9c5545f54fb3fcc815f52b8a7cf.
```

Testcase that should still break the build (because line is longer than 64 chars before the commit hash):
```
GrpcService: fix some logging nits

These mistakes were made in this GrpcService's RunIntoMeService commit: 45faeca2f0e7c9c5545f54fb3fcc815f52b8a7cf.
```

realmarv:
@webwarrior-ws please work on this issue if you are taskless.

knocte:
@realmarv I think @webwarrior-ws got stuck with this task, I don't remember his reasoning. Please pair with him on it after he's done with his NLitecoin low-prio task.

PR #133, opened 2023-08-01
Body soft max line length exception

webwarrior-ws:
Add tests for case described in https://github.com/nblockchain/conventions/issues/124.
Add exception to bodySoftMaxLineLength function when line exceeds max length but ends with a git commit hash.

Fixes https://github.com/nblockchain/conventions/issues/124

### DIFF
--- a/commitlint/plugins.ts
+++ b/commitlint/plugins.ts
@@ -399,7 +399,21 @@ export abstract class Plugins {
 
                         let lineIsFooterNote = Helpers.isFooterNote(line);
 
-                        if (!isUrl && !lineIsFooterNote) {
+                        let commitHashPattern = `([0-9a-f]{40})`;
+                        let anySinglePunctuationCharOrNothing = `[\.\,\:\;\?\!]?`;
+                        let index = line.search(
+                            commitHashPattern +
+                                anySinglePunctuationCharOrNothing +
+                                `$`
+                        );
+                        let endsWithCommitHashButRestIsNotTooLong =
+                            index != -1 && index < bodyMaxLineLength;
+
+                        if (
+                            !isUrl &&
+                            !lineIsFooterNote &&
+                            !endsWithCommitHashButRestIsNotTooLong
+                        ) {
                             offence = true;
                             break;
                         }
--- a/commitlint/plugins.test.ts
+++ b/commitlint/plugins.test.ts
@@ -323,6 +323,24 @@ test("body-max-line-length8", () => {
     expect(bodyMaxLineLength8.status).toBe(1);
 });
 
+test("body-max-line-length9", () => {
+    // see https://github.com/nblockchain/conventions/issues/124
+    let commitMsgWithLargeBody =
+        "GrpcService: fix some logging nits\n\n" +
+        "These mistakes were made in 45faeca2f0e7c9c5545f54fb3fcc815f52b8a7cf.";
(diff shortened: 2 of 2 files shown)


knocte:
@webwarrior-ws CI is red

commitlint/plugins.ts
```diff
@@ -399,7 +399,20 @@ export abstract class Plugins {
 
                         let lineIsFooterNote = Helpers.isFooterNote(line);
 
-                        if (!isUrl && !lineIsFooterNote) {
+                        let endsWithCommitHashButRestIsNotTooLong = false;
+                        let commitHashAtTheEndMatch =
+                            line.match(`([0-9a-f]{40})\.?$`);
```
knocte, line 404: @webwarrior-ws let's make this pattern match not only when a dot is at the end, but also any punctuation sign (comma, colon, semicolon) and even no character.

knocte: (also exclamation and interrogation: ! ?)

commitlint/plugins.ts
```diff
@@ -399,7 +399,20 @@ export abstract class Plugins {
...
+                        let index = commitHashAtTheEndMatch?.index;
+                        if (index !== undefined) {
```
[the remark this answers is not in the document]
webwarrior-ws: Match object can be `null` (if there is no match). Initially I checked it for null, and if not null, compared index. But compiler didn't like it for some reason, so we settled on this solution.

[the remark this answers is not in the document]
knocte: what's the compiler error?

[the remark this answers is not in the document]
knocte: or maybe use the search() function and then just check for null instead of checking index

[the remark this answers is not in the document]
webwarrior-ws: Replaced `match()` with `search()`; that also simplified code.

commitlint/plugins.ts
webwarrior-ws: Included other punctuation signs in the regex.

knocte:
@webwarrior-ws the comments I put before about punctuation chars is not yet addressed.

knocte:
@realmarv why didn't CI fail in this job (check commits 1by1)?: https://github.com/nblockchain/conventions/actions/runs/5736837375/job/15547571374  I don't understand

commitlint/plugins.ts
webwarrior-ws: Oops, I've accidentally overwritten my changes to regex when replacing match() with search(). Now it is fixed.

webwarrior-ws:
> @realmarv why didn't CI fail in this job (check commits 1by1)?: https://github.com/nblockchain/conventions/actions/runs/5736837375/job/15547571374 I don't understand

I put "failing test" into commit message of commit that adds tests.

commitlint/plugins.ts
```diff
@@ -399,7 +399,15 @@ export abstract class Plugins {
 
                         let lineIsFooterNote = Helpers.isFooterNote(line);
 
-                        if (!isUrl && !lineIsFooterNote) {
+                        let index = line.search(`([0-9a-f]{40})[\.\,\:\;\?\!]?$`);
```
knocte, line 402: @webwarrior-ws let's split this pattern in two variables please, to make it more readable, one called commitHashPattern and the other anySinglePunctuationCharOrNothing

webwarrior-ws: Split into 2 variables.

Issue #134, opened 2023-08-08
body-prose false positive


knocte:
https://github.com/nblockchain/geewallet/actions/runs/5796723490/job/15711151832 from https://github.com/nblockchain/geewallet/commit/6b3c64ed493e9481560e96e1858a938aaa1b56e7:

```
Backend: adjust legacy obj path (#218)

Using the same folder for two projects (e.g.
GWallet.Backend.fsproj and GWallet.Backend-legacy.fsproj) that
use different frameworks (e.g. NetStandard2.0 one and .NET4.x
one), causes problems due to incompatible caches in obj folder,
so by adjusting the cache obj folder for the legacy version we
workaround this problem.
```

Result: 
```
commitlint/cli@17.6.7
+ npx commitlint --from HEAD~1 --to HEAD --verbose
⧗   input: Backend: adjust legacy obj path (#218)

Using the same folder for two projects (e.g.
GWallet.Backend.fsproj and GWallet.Backend-legacy.fsproj) that
use different frameworks (e.g. NetStandard2.0 one and .NET4.x
one), causes problems due to incompatible caches in obj folder,
so by adjusting the cache obj folder for the legacy version we
workaround this problem.
✖   The body of the commit message (as opposed to the commit message title) is composed of paragraphs. Please begin each paragraph with an uppercase letter and end it with a dot (or other valid character to finish a paragraph).
For reference, these are the guidelines that include our commit message conventions: https://github.com/nblockchain/conventions/blob/master/WorkflowGuidelines.md [body-prose]
```

realmarv:
It probably isn't related to this issue but the commit message you mentioned should fail because of "body-paragraph-line-min-length" rule.

knocte:
> It probably isn't related to this issue but the commit message you mentioned should fail because of "body-paragraph-line-min-length" rule.

I disagree, and the rule agrees with me. Remember how the rule works: if next word is too long, then it cannot complain because the next word could not be moved to the previous line.

PR #136, opened 2023-08-14
Fix body prose false positive

Mersho:
The `replace()` method does not replace all. Use `replaceall()` to resolve issue #134 .


knocte:
Amazing investigative work! Will review tomorrow.

PR #137, opened 2023-08-15
commitlint/helpers.ts: fix errMessageSuffix URL

Mersho:


knocte:
Why is PR title different than commit message title?

knocte:
> commitlint/helpers: fix errMessageSuffix url

Nit: URL is an acronym, so please capitalize it.

PR #142, opened 2023-09-18
commitlint: build in strict mode

Mersho:


aarani:
What's the plan here? Just adding all strict flags in TS compiler? Sorry but I hate it! This is not a spacecraft control panel. /cc @knocte

knocte:
> What's the plan here? Just adding all strict flags in TS compiler? Sorry but I hate it! This is not a spacecraft control panel.

The plan is this: https://github.com/nblockchain/lnp2pbot/issues/2 (inspired by this: https://hackernoon.com/making-typescript-truly-strongly-typed )

knocte:
> One way to fix this is to assign a function expression to a variable name instead:

AFAIU, your commit doesn't do this anymore.

knocte:
@Mersho can I get an update here, what happened with this PR?

Mersho:
> @Mersho can I get an update here, what happened with this PR?

CI of the last commit failed on "installing dependencies".
Also, the CI of the second commit failed but we don't have the logs. I'm going to investigate it now.

Mersho:
My branch CI is green but if you want a green PR CI we need to add 'failing test' to first commit:

`please make sure that the commit message contains the 'failing test' term`

knocte:
> My branch CI is green but if you want a green PR CI we need to add 'failing test' to first commit:

OR HOW ABOUT FIXING THE CI OF FIRST COMMIT INSTEAD OF HAVING A FAKE COMMIT MESSAGE?????

knocte:
>> The error below occurs when a function declaration is inside a
block in strict mode when targeting ‘ES3’ or ‘ES5’. One
way to fix this is to assign a function expression to a
variable name instead:
> commitlint/plugins.ts:19:26 - error TS1251: Function declarations are not allowed inside blocks

I disagree with the way you are fixing this!!! THE POINT OF USING THE STRICT FLAG IS TO BE MORE STRICT! **If you add the strict flag but later add another flag that makes the strict flag weaker, WHAT IS THE POINT????** Just fix the goddamn compiler error!!!!!!!

knocte:
@Mersho can you rebase this on Monday please

knocte:
> Also print the typescript version on the CI as well.

When you have to write the word "also" in a commit message, you're already suspicious of mixing things that maybe shouldn't be mixed.

And in this case, in fact, I'm going to ask you to split that commit in two, and separate the devDependencies part to another PR please.

tsconfig.json
```diff
@@ -2,6 +2,7 @@
     "compilerOptions": {
         "lib": ["ES2021"],
         // TODO: explore replacing the below setting with just target:ES2015
-        "downlevelIteration": true
-    }
+        "downlevelIteration": true,
+        "strict": true,
```
knocte, line 6: @Mersho please:
1. move this to be the very first flag of the JSON file
2. I have NEVER told you to move the `tsc --version` outside of this PR. READ MY LATEST MESSAGES AGAIN TO TRY TO UNDERSTAND WHAT I SAID

tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
     "compilerOptions": {
+        "strict": true,
         "lib": ["ES2021"],
         // TODO: explore replacing the below setting with just target:ES2015
-        "downlevelIteration": true
-    }
+        "downlevelIteration": true,
+    },
```
knocte, line 7: @Mersho these 2 lines don't need to change

PR #149, opened 2024-01-18
CI: add compile typescript step

webwarrior-ws:
Add "compile typescript" step to CI so that any compiler errors in tests are reported rather than lead to test failure with unclear cause.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -102,9 +102,7 @@ jobs:
       - name: Run typescript compiler
         # This step is needed because jest would swallow any compiler error
         # and just declare test as failed without any explanation.
-        run: |
-          echo "{}" > ./tsconfig.json
-          npx tsc
+        run: npx tsc
       - name: Run tests to validate our plugins
         run: yarn jest
 
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -99,6 +99,12 @@ jobs:
           npm --version
           npx commitlint --version
           yarn --version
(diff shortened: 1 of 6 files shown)


.github/workflows/CI.yml
```diff
@@ -99,6 +99,11 @@ jobs:
           npm --version
           npx commitlint --version
           yarn --version
+          npx tsc --version
+      - name: Run typescript compiler
+        # This step is needed because jest will swallow any compiler error
```
knocte, line 104: @webwarrior-ws shouldn't this say "would" instead of "will"?

webwarrior-ws: It will swallow a compiler error, if there is any.

knocte: yes but after this commit it will not; so putting this comment is misleading, change the "will" with "would" please

webwarrior-ws: Done.

knocte:
I just compared the compiler errors between the 1st and 2nd commit and I don't see any differences; so is the commit title "CI: set typescript version to 4.9" really needed at all?

webwarrior-ws:
> I just compared the compiler errors between the 1st and 2nd commit and I don't see any differences; so is the commit title "CI: set typescript version to 4.9" really needed at all?

You're right, this step is not needed. Removed that commit.

knocte:
Please split the 1st commit in two. One that simply adds the compilation, and a 2nd commit that explains the downleveliteration thing (also, this latter setting can be set in the json file rather than an argument).

PR #152, opened 2024-01-22
Fix for Issue #148

Mersho:
Fixes https://github.com/nblockchain/conventions/issues/148

### DIFF
--- a/commitlint/plugins.test.ts
+++ b/commitlint/plugins.test.ts
@@ -778,6 +778,24 @@ test("footer-refs-validity8", () => {
     expect(footerRefsValidity8.status).toBe(0);
 });
 
+// This test reflects this issue: https://github.com/nblockchain/conventions/issues/148
+test("footer-refs-validity9", () => {
+    let commitMsgWithTwoCodeBlocksAtBodyWithRef =
+        "foo: blah blah" +
+        "\n\n" +
+        "Blah blah [1]:" +
+        "\n\n" +
+        "```\nsomeCodeBlock\n```" +
+        "\n\n" +
+        "[1] Stack trace:" +
+        "\n\n" +
+        "```\nsomeCodeBlock\n```";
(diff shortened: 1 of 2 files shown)


commitlint/plugins.test.ts
```diff
@@ -778,6 +778,30 @@ test("footer-refs-validity8", () => {
...
+        "\n\n" +
+        "```\nsomeCodeBlock\n```" +
+        "\n\n" +
+        "Blah blah blah: [1]." +
+        "\n\n" +
+        "[1] someUrl://blahblah.com" +
+        "\n" +
+        "[2] Stack trace:" +
+        "\n\n" +
+        "```\nsomeCodeBlock\n```";
```
knocte, line 798: @Mersho now that you understand the issue better (because you have found a fix), cannot you simplify this testcase? I mean, the first thing I see that could be simplified is that the latter code block is not needed to reproduce the bug, right?

Mersho: yes, you're right.

Mersho:
We might have a single backtick code block.
i can write a testcase but it's not related to this PR and issue.

knocte:
Single backticks are not meant for code blocks.

knocte:
This PR is wrong now, because CI of first commit is not failing.

Mersho:
i guess we need the latter code block, let me put it back.

knocte:
I guess? Then you don't really understand the bug yet.

knocte:
> This commit addresses an issue where the removeAllCodeBlocks
function was inadvertently removing all text following a code
block after replacing it. This fix ensures that only the code
blocks are replaced, and the rest of the text remains intact.

I told you yesterday to REWRITE this explanation to cover your NEW understanding of the bug (the one that made us UNDERSTAND that we need 2 code blocks in the minimal testcase).

knocte:
Missing "Fixes" sentence in commit msg.

PR #170, opened 2024-03-14
CI: configure git inside docker container

Mersho:
Commitlint fails to execute because a Git error is occurring and it cannot recognize the Git repository:

```
Error: fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

After investigating, I found out that when we run the commitlint plugins tests job on a Docker container, there’s no `.git` directory at all.

Also, when `tsc` is executed, `commitlint.config.js` will be created and commitlint mistakenly reads that file instead of `commitlint.config.ts`. It is always better to use tsconfig `outDir` flag to prevent this issues.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -106,14 +106,11 @@ jobs:
     needs: build-ts
     # TODO: upgrade to 24.04->24.04 (because LTS) as soon as it's available, instead of 22.04->23.10
     runs-on: ubuntu-22.04
-    container:
-      image: "ubuntu:23.10"
     steps:
       - uses: actions/checkout@v4
       - name: Install required dependencies
         run: |
-          apt update
-          apt install --yes sudo
+          sudo apt update
           sudo apt install --yes --no-install-recommends git ca-certificates
 
           sudo apt install --yes --no-install-recommends npm curl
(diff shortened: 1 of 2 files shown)


Mersho:
I tried to pass `commitlint.config.ts` through the --config flag (instead of removing it), but it’s not working on my local.

Mersho:
> IT'S NOT WORKING" AGAIN!

BECAUSE I DON'T KNOW THE REASON. It just keeps reading commitlint.config.js again!

knocte:
> BECAUSE I DON'T KNOW THE REASON. It just keeps reading commitlint.config.js again!

IT IS NOT ABOUT NOT KNOWING THE REASON! SAYING "NOT WORKING" WITHOUT SHARING MORE DETAILS WILL ALWAYS REQUIRE AN EXTRA FEEDBACK LOOP ASKING YOU WHY DO YOU THINK IT'S NOT WORKING! WHAT ERROR DID YOU SEE? WHAT DID YOU EXPERIENCE IN ORDER TO THINK IT'S NOT WORKING? IT'S APPALING, WE'VE TALKED ABOUT THIS SO MANY TIMES.

PR #171, opened 2024-03-20
CI,commitlint(tests): upgrade to latest commitlint

webwarrior-ws:
Convert commitlint plugins to ES module so that it can be loaded with "module": "node16" setting that's needed to load commitlint, which is ES module now.

Move tests to separate directory so they are not considered part of "commitlint" module, because jest's support for ES modules is still experimental.

Run tsc for tests separately and have separate tsconfig.json for tests directory as tests use CommonJS modules which are incompatible with ES modules.


knocte:
Can this work be separated into 2 commits? One that moves the things first to a different folder, and the 2nd one that upgrades commitlint.

commitlint/tests/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+    "compilerOptions": {
+        "strict": true,
+        "lib": ["ES2021"],
+        // TODO: explore replacing the below setting with just target:ES2015
+        "downlevelIteration": true,
+        "types": ["node", "jest"],
+    }
+}
```
knocte, line 9: @webwarrior-ws isn't there a way to have a single tsconfig file for the whole repo? to respect DRY

package.json
```diff
@@ -1,5 +1,15 @@
 {
+    "type": "module",
     "dependencies": {
-      "cosmiconfig": "8.0.0"
+        "@commitlint/config-conventional": "^19.1.0",
+        "@commitlint/types": "^19.0.3",
```
knocte, line 5: @webwarrior-ws by inserting these 2 lines above, you're contradicting what the file `version.config` says

commitlint/tests/tsconfig.json
webwarrior-ws: We need different settings for tests, as they use CommonJS modules and not ES modules.

knocte:
> upgrade to latest commitlint
>
> Upgrade to latest commitlint version.

Why do you duplicate this text (from the title) in the commit msg body?

commitlint/tests/tsconfig.json
knocte: Ok, what would be the best way to fix this? Maybe have a 3rd commit in this PR that migrates the tests from jest to yarn? I only suggested the latter because it seems that it is `yarn test` what commitlint upstream uses to run their tests, but correct me if I'm wrong.

webwarrior-ws: AFAIK yarn is just a runner. commitlint uses `vitest`:
https://github.com/conventional-changelog/commitlint/blob/e41023d01c3f8569c3b91f4362df9bdc645b5f3f/package.json#L25

webwarrior-ws: Added a commit that migrates from jest to vitest

.github/workflows/CI.yml
```diff
@@ -89,8 +89,8 @@ jobs:
           npm --version
       - name: Install yarn
         run: |
-          npm install --verboase --global yarn
```
knocte, line 92: @webwarrior-ws oh wow this is an unfortunate typo; do you mind fixing this as a separate commit please (this way, in case we need to revert this commit later, we would not bring back a typo)

.github/workflows/CI.yml
```diff
@@ -122,13 +122,14 @@ jobs:
...
-          yarn add --dev jest typescript ts-jest @types/jest ts-node
+          yarn add --dev typescript ts-node
       - name: Install commitlint dependencies
         run: |
           . commitlint/version.config
           npm install --verbose @commitlint/types@$COMMITLINT_VERSION
       - name: Run typescript compiler
-        run: npx tsc
+        run: |
+          npx tsc
```
knocte, line 132: @webwarrior-ws unneeded change?

.github/workflows/CI.yml
```diff
@@ -139,7 +140,10 @@ jobs:
       - name: Install commitlint
         run: ./commitlint.sh --version
       - name: Run tests to validate our plugins
-        run: yarn jest --config jest.config.ts
+        run: |
+          # remove *.js files so that the same tests don't run twice
+          rm ./commitlint/tests/*.js
```
knocte, line 145: @webwarrior-ws I believe this is a problem in master branch too so let's put this as a separate commit before this commit

.github/workflows/CI.yml
webwarrior-ws: Moved to separate commit

.github/workflows/CI.yml
webwarrior-ws: Reverted back to single-line command

.github/workflows/CI.yml
webwarrior-ws: Moved to separate commit

PR #172, opened 2024-08-30
commitlint: honor our own WorkflowGuidelines

knocte:
The workflow guidelines state that we should avoid the keyword "undefined". This is also extra good because it seems that the ? guards against both null and undefined, at least according to this example:

https://github.com/conventional-changelog/commitlint/commit/7ab4bab31f4b19ddedd850e435398037437007b6

### DIFF
--- a/commitlint/helpers.ts
+++ b/commitlint/helpers.ts
@@ -20,11 +20,8 @@ export abstract class Helpers {
         potentialString: any,
         paramName: string
     ): string | null {
-        if (potentialString === null || potentialString === undefined) {
-            // otherwise, String(null) might give us the stupid string "null"
-            return null;
-        }
-        return String(potentialString);
+        // this null/undefined check is required, otherwise, String(null) might give us the stupid string "null"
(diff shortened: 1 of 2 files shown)


knocte:
@webwarrior-ws on Monday can you please verify that this PR is correct? By this, I mean that your review should also make sure that what I'm saying in the commit msg is correct, i.e. does the '?' char guard against both null and undefined?

knocte:
@webwarrior-ws and please investigate why this PR didn't trigger a CI and why the schedule trigger of this CI workflow hasn't worked for the last 3 months?

knocte:
> @webwarrior-ws and please investigate why this PR didn't trigger a CI and why the schedule trigger of this CI workflow hasn't worked for the last 3 months?

Nevermind. I have no idea how but the workflow was disabled, I had to re-enable it in the GitHub UI, Actions tab.

knocte:
@webwarrior-ws nevermind, had some extra time today and I checked myself.

PR #174, opened 2024-09-05
commitlint,docs: start using FP patterns

knocte:


knocte:
@webwarrior-ws hey Taras can you investigate why the 2nd commit of this PR breaks tests please? I'm very confused.

webwarrior-ws:
You've used class `None` instead of its instance, and that lead to `instanceof None` check to always fail.
I've added commit to fix it: https://github.com/webwarrior-ws/conventions/commits/wip/fp/

knocte:
Thanks

webwarrior-ws:
There was a bug in `IsInstanceOf`, I fixed it - see last commit in https://github.com/webwarrior-ws/conventions/commits/wip/fp/
CI is red because of formatting in other files.

PR #176, opened 2024-11-05
scripts: fix unpinnedNugetPackage... script

webwarrior-ws:
Use working dir instead of script's parent directory in unpinnedNugetPackageReferenceVersionsInProjects.fsx to search for projects as it is intended behavior.

### DIFF
--- a/scripts/unpinnedNugetPackageReferenceVersionsInProjects.fsx
+++ b/scripts/unpinnedNugetPackageReferenceVersionsInProjects.fsx
@@ -9,10 +9,13 @@ open System.IO
 #load "../src/FileConventions/Helpers.fs"
 
 let rootDir = Path.Combine(__SOURCE_DIRECTORY__, "..") |> DirectoryInfo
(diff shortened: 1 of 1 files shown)


knocte:
I don't see any other scripts doing this, how can they be working?

webwarrior-ws:
Other scripts do use current directory, some also check script's parent dir.

knocte:
> Other scripts do use current directory,

Which ones?

knocte:
Actually, this is the approach that needs to be used: abfde9af99aaedda4a6348c9ae916a4732de6bcb

webwarrior-ws:
I don't understand logic of `PreferLessDeeplyNestedDir`. We use current working dir, but if it has longer way to root use script's parent dir? Why would that be useful?

knocte:
Because $SOMEPATH/externalRepo is always less nested than $SOMEPATH/externalRepo/conventions/scripts

PR #180, opened 2024-11-21
GithubCI: use Ubuntu 24.04 instead of 23.10

webwarrior-ws:
Because 23.10 is no longer supported and 24.04 is LTS release.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -9,13 +9,16 @@ on:
   schedule:
     - cron: "0 0 * * *"
 
+# As we use .NET 8 now, to run tools compiled for .NET 6 we need to set roll-forward to major
+env:
+  DOTNET_ROLL_FORWARD: Major
+
 jobs:
   build-fs:
     name: Build F#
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -32,6 +31,10 @@ jobs:
           # The remote certificate is invalid because of errors in the certificate chain: UntrustedRoot
           apt install --yes --no-install-recommends ca-certificates
 
+          # register backports repository (needed for dotnet6)
+          sudo apt install  --yes --no-install-recommends software-properties-common
+          sudo add-apt-repository ppa:dotnet/backports
```
knocte, line 36: @webwarrior-ws this is unacceptable, we don't want to use PPAs

webwarrior-ws: How are we going to install .NET 6 then?

knocte: well, if dotnet6 is not available in 24.04 then we upgrade

webwarrior-ws: Upgraded to .NET 8

PR #181, opened 2025-03-11
docs: expanded bad practices section

webwarrior-ws:
Added point about avoiding usage of JavaScript/TypeScript spread operator (`...`).

### DIFF
--- a/docs/WorkflowGuidelines.md
+++ b/docs/WorkflowGuidelines.md
@@ -249,6 +249,18 @@
             someVar = expression;
         }
         ```
+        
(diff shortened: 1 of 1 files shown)


knocte:
In all of the previous cases in the documentation about bad operators, it is explained why. You forgot to explain why.

webwarrior-ws:
> In all of the previous cases in the documentation about bad operators, it is explained why. You forgot to explain why.

I personally don't see why it's bad. So please explain why you don't like it and I'll add that explanation to the documentation.

PR #182, opened 2025-04-08
scripts: combined inconsistent versions script

webwarrior-ws:
Added a script for finding inconsistent package versions across both nuget references in projects and .fsx script files.

Refactored scripts for checking versions in project files and .fsx files to be able to reuse existing code for new script.


knocte:
* CI is red.
* Please add a GitHub step that invokes this script.
* Please add a new commit, in between the refactoring and the last one, that adds unit tests (in the same way that there are unit tests for the functionality of the F# script that detects inconsistent nuget versions in .fsx files). Don't overengineer this, just very very simple unit test that has inconsitent version across a .fsx file and a .fsproj file.

webwarrior-ws:
> * Please add a GitHub step that invokes this script.
This will be a false positive because of all the test files (DummyFiles dir).

webwarrior-ws:
Added a test

knocte:
Shouldn't CI fail because Tests commit is red and it doesn't contain the term "failing test" in the commit msg? please investigate.

knocte:
> Shouldn't CI fail because Tests commit is red and it doesn't contain the term "failing test" in the commit msg? please investigate.

I'm referring to scripts/checkCommits1by1.fsx btw. It should fail with the way you have titled commits at the moment, let's find why it didn't fail.

webwarrior-ws:
> Shouldn't CI fail because Tests commit is red and it doesn't contain the term "failing test" in the commit msg? please investigate.

But it does have "Contains failing test." at the end

knocte:
> But it does have "Contains failing test." at the end

Ah missed that, I thought we needed it in the title. Ok, merging.

knocte:
> You didn't do this! sigh, please create new PR for this.

Nevermind, done in https://github.com/nblockchain/conventions/pull/183

PR #185, opened 2025-04-15
scripts,FileConventions: allow specifying dirs

webwarrior-ws:
To search for .fsx and project files in combined versions check script. If no arguments are passed, current working directory is used.


knocte:
@webwarrior-ws pls update ReadMe.md with new script file names.

Issue #117, opened 2023-07-05
wrapLatestCommitMsg script failed ignoring a code block


tehraninasab:
The script doesn't ignore the code block in the following commit message:

````
scripts/styleChecker.fsx: add

Add styleChecker.fsx script to check style of our F#, TS
and YML codes.

The `git restore package.json` command in the
styleChecker.fsx script was failing with the following
error, even when I was running the
`git config --global --add safe.directory '*'` command in
both the styleChecker.fsx script and in the CI. In the
issue [1], it's suggested by someone to use --system
instead of --global in the mentioned command, when the
git command is running in a container, which solved the
problem.

```
fatal: detected dubious ownership in repository at '/__w/conventions/conventions'
To add an exception for this directory, call:

	git config --global --add safe.directory /__w/conventions/conventions

Error when running 'git restore package.json'
Fsdk.Process+ProcessFailed: Exception of type 'Fsdk.Process+ProcessFailed' was thrown.
   at Fsdk.Process.ProcessResult.Unwrap(String errMsg)
   at Fsdk.Process.ProcessResult.UnwrapDefault()
   at FSI_0002.RunPrettier(String arguments)
   at <StartupCode$FSI_0002>.$FSI_0002.main@()
Stopped due to error
Error: Process completed with exit code 1.
```

[1] https://github.com/actions/checkout/issues/1048
````

PR #135, opened 2023-08-09
FileConventions: fix WrapText function 

parhamsaremi:
Ignore codeblocks while splitting the text into paragraphs because there might be a codeblock with multiple paragraphs.

Fixes https://github.com/nblockchain/conventions/issues/117

PR #187, opened 2025-04-29
FileConventions: fix WrapText function

webwarrior-ws:
Ignore codeblocks while splitting the text into paragraphs because there might be a codeblock with multiple paragraphs.

Fixes https://github.com/nblockchain/conventions/issues/117

Supersedes https://github.com/nblockchain/conventions/pull/135

### DIFF
--- a/src/FileConventions.Test/WrapTextTests.fs
+++ b/src/FileConventions.Test/WrapTextTests.fs
@@ -107,3 +107,25 @@ let WrapTextTest5() =
     let expectedResult = text
 
     Assert.That(WrapText text characterCount, Is.EqualTo expectedResult)
+
+[<Test>]
+let WrapTextTest6() =
+    let characterCount = 64
+
+    let commitMsg =
+        "foo: this is a header"
+        + Environment.NewLine
+        + Environment.NewLine
+        + "This is a body:"
+        + Environment.NewLine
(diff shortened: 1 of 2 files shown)


webwarrior-ws:
Looks OK, but the fact that `currentParagraph` and `paragraphs` are both `List<string>` is a bit confusing.

knocte:
> Looks OK, but the fact that currentParagraph and paragraphs are both List<string> is a bit confusing.

You mean paragraphs should be List<Paragraph> and then make Paragraph be an alias of List<string>? Don't just critize, offer solution.

webwarrior-ws:
> > Looks OK, but the fact that currentParagraph and paragraphs are both List is a bit confusing.
> 
> You mean paragraphs should be List and then make Paragraph be an alias of List? Don't just critize, offer solution.

I get why they are the same type. I'm thinking more of renaming. For example `currentParagraph ` -> `currentParagraphLines`

knocte:
Sure let's add an extra commit to rename.

knocte:
> You mean paragraphs should be List<Paragraph> and then make Paragraph be an alias of List<string>? Don't just critize, offer solution.

FYI I hadn't escaped my comment properly so I guess it was confusing. I edited my comment now.

Issue #173, opened 2024-08-30
Adopt typescript-eslint


knocte:
Example: https://github.com/conventional-changelog/commitlint/commit/d5d572d5f0fe0138127f943937e7ff13232e2eb8

webwarrior-ws:
Should I use default `eslint.config.mjs` (as in https://typescript-eslint.io/getting-started) or the one from commit that you've given as example?

knocte:
Default for now.

PR #194, opened 2025-05-01
GithubCI: install and run typescript-eslint

webwarrior-ws:
Installed typescript-eslint and added CI step to run it.

Ran `npx eslint . --fix`. It replaced `let` declarations with `const` ones when value is not reassigned.

Fixed some other eslint warnings. Silenced remaining warnings.

Fixed type error that appeared because of using typed imports in testHelpers.ts.

Fixes https://github.com/nblockchain/conventions/issues/173

### DIFF
--- a/commitlint/fpHelpers.ts
+++ b/commitlint/fpHelpers.ts
@@ -51,11 +51,13 @@ export class OptionStatic {
 }
 
 export class TypeHelpers {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any -- method accepts any value by design
     public static IsNullOrUndefined(variable: any) {
         return variable === null || variable === undefined;
     }
 
     // because instanceof doesn't work with primitive types (e.g. String), taken from https://stackoverflow.com/a/58184883/544947
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any -- method accepts any value by design
     public static IsInstanceOf(variable: any, type: any) {
         if (TypeHelpers.IsNullOrUndefined(variable)) {
             throw new Error(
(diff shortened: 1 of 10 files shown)


commitlint/fpHelpers.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint @typescript-eslint/no-explicit-any: "off" */
```
knocte, line 1: @webwarrior-ws why? unacceptable, let's fix it

webwarrior-ws: `public static IsNullOrUndefined(variable: any)`
`public static IsInstanceOf(variable: any, type: any)`
How do you propose to fix it? Are these methods not expecting literally `any` value?

knocte: ah shit, didn't realise about this; can the no-explicit-any:off be applied per func instead of per file?

webwarrior-ws: > ah shit, didn't realise about this; can the no-explicit-any:off be applied per func instead of per file?

It can be applied for the next line.

knocte: Then let's do that

webwarrior-ws: Already switched to using `eslint-disable-next-line` in `fpHelpers.ts` and `helpers.ts`.
For `commitlint.config.ts` I left original `/* eslint @typescript-eslint/no-explicit-any: "off" */` annotation because there are many places that use `any`.

knocte:
> Fixed some other eslint warnings. Silenced remaining warnings.
> Fixed type error that appeared because of using typed imports
in testHelpers.ts.

You're again writing a list of things in a paragraph, as prose. A list of things is a list! List the items in a bullet list!

knocte:
CI is red.

PR #198, opened 2025-07-28
GithubCI: upgrade FSharpLint to latest version

webwarrior-ws:
- Upgraded FSharpLint to version 0.26.1--date20250724-0123.git-3924085
- Applied FSharpLint suggestions

### DIFF
--- a/fsharplint.json
+++ b/fsharplint.json
@@ -287,6 +287,7 @@
     "uselessBinding": { "enabled": true },
     "tupleOfWildcards": { "enabled": true },
     "favourTypedIgnore": { "enabled": true },
+    "favourNonMutablePropertyInitialization": { "enabled": false },    
     "favourReRaise": { "enabled": true },
     "favourStaticEmptyFields": { "enabled": true },
     "favourConsistentThis": {
@@ -331,6 +332,7 @@
             "additionalPartials": []
         }
     },
+    "ensureTailCallDiagnosticsInRecursiveFunctions": { "enabled": false },
     "hints": {
         "add": [
             "not (a =  b) ===> a <> b",
--- a/src/FileConventions/NugetVersionsCheck.fs
+++ b/src/FileConventions/NugetVersionsCheck.fs
@@ -34,7 +34,7 @@ module MapHelper =
                     yield theKey, valsForKey
             }
 
-        keyValuePairs |> Map.ofSeq
+        Map.ofSeq keyValuePairs
 
     let MergeMaps<'Key, 'Value when 'Key: comparison and 'Value: comparison>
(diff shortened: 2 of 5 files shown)


src/FileConventions/NugetVersionsCheck.fs
```diff
@@ -170,10 +170,10 @@ let private notSubmodule
 
     not(
         getSubmoduleDirsForThisRepo()
-            .Any(fun d -> dir.FullName = d.FullName)
+            .Any(fun each -> dir.FullName = each.FullName)
```
knocte, line 173: @webwarrior-ws each is a bad name, use submodule?

knocte: rather, submoduleDir

knocte:
CI is red. Also, shouldn't the fsharp-lint.json file be augmented?

knocte:
> Contains failing CI

This text is confusing to a potential future developer.

webwarrior-ws:
> > Contains failing CI
> 
> This text is confusing to a potential future developer.

Clarified what causes failing CI in commit message.

knocte:
I know, but a commit doesn't conain any CI, that's the wrong verb to use.

knocte:
> Why? Do you have some rules or rule settings in mind that need to be adjusted?

No. But between the old and the new version of FSharpLint being upgraded, there are more rules, so then the config file needs to be updated.

webwarrior-ws:
> > Why? Do you have some rules or rule settings in mind that need to be adjusted?
> 
> No. But between the old and the new version of FSharpLint being upgraded, there are more rules, so then the config file needs to be updated.

Unless I'm missing something, that is not required. If something is not specified in the config, default is used.

knocte:
It's better to be explicit about what *default* means, and we have had the convention of doing this for every FSharpLint upgrade, so let's keep doing it.

knocte:
1st commit should be squashed with last

fsharplint.json
```diff
@@ -331,6 +331,10 @@
             "additionalPartials": []
         }
     },
+    "unneededRecKeyword": { "enabled": true },
+    "ensureTailCallDiagnosticsInRecursiveFunctions": { "enabled": false },
+    "favourNonMutablePropertyInitialization": { "enabled": true },
+
```
knocte, line 337: @webwarrior-ws does config file from upstream FSharpLint repo contain this extra EOL?

webwarrior-ws:
> > let's mimic as close as possible the upstream one
> 
> When you looked at the file from upstream, you completely ignored my sentence above ^

Done

PR #202, opened 2025-08-06
GitHubCI,scripts: new script for fantomless/fantomas

knocte:
This way we can run it outside this repo.


knocte:
TODO: make sure this fixes https://github.com/tarsgate/conventions/issues/249

PR #206, opened 2025-11-03
FileConventions: skip values not in env dict

webwarrior-ws:
When searching for vars referenced in Github workflow files. This is needed because workflows in pulumi-deploy now load vars from .env file and so there can be references to variables that are not in env dictionary of .yaml file.


knocte:
Typo: Comventions

PR #207, opened 2025-11-06
Add "dynamic uses" action

webwarrior-ws:
To the repository root so it can be referenced and used in Github workflows. The action is originally from [1] and is
a workaround to problem outlined in [2].

Use "dynamic uses" action in workflow to specify version of actions/checkout only once and avoid duplication.

[1] https://github.com/jenseng/dynamic-uses
[2] https://github.com/orgs/community/discussions/25246

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -22,7 +22,7 @@ jobs:
     container:
       image: "ubuntu:24.04"
     steps:
-      - uses: webwarrior-ws/conventions@dynamic-uses-action
+      - uses: nblockchain/conventions@main
         with:
           uses: actions/checkout@v${{ env.CHECKOUT_ACTION_VERSION }}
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -21,7 +22,9 @@ jobs:
     container:
       image: "ubuntu:24.04"
     steps:
-      - uses: actions/checkout@v4
+      - uses: webwarrior-ws/conventions@dynamic-uses-action
```
knocte: @webwarrior-ws cannot this reference itself with a relative path?

webwarrior-ws:
> Can you rename action.yml to dynamc-uses-action.yml and place it inside .github dir? Would that work?

That would work for this repo but I don't think such action can be referenced from other repository.

knocte:
Ok let's see, merging.

PR #213, opened 2025-12-17
scripts/deleteOldArtifacts.fsx: fetch all commits

webwarrior-ws:
Make sure git has complete history before checking for branchless commits. Otherwise some commits may be marked as branchless even if they are not because local repo doesn't have them.

### DIFF
--- a/scripts/deleteOldArtifacts.fsx
+++ b/scripts/deleteOldArtifacts.fsx
@@ -251,6 +251,22 @@ match whatArtifactsToDelete with
     Console.WriteLine
         "About to delete all the artifacts of branchless commits..."
 
+    // Make sure git has complete history before checking for branchless commits.
+    let gitFetchUnshallowResult =
+        Fsdk.Process.Execute(
+            {
+                Command = "git"
+                // The special depth 2147483647 means infinite depth, see https://git-scm.com/docs/shallow
+                Arguments = "fetch --depth=2147483647"
+            },
+            Echo.Off
+        )
+    // Fsdk may categorize normal output of git fetch as WarningsOrAmbiguous, so have to treat it as succcess.
+    match gitFetchUnshallowResult.Result with
+    | Success _
+    | WarningsOrAmbiguous _ -> ()
+    | Error _ -> gitFetchUnshallowResult.UnwrapDefault() |> ignore<string>
+
     let orphanArtifactIdsToDelete =
         artifactIds
         |> Seq.filter(fun item ->


knocte:
Tested?

webwarrior-ws:
> > There was likely a reason
> 
> Also, remember that we try to always to add reasons to commit msgs in this company, so you could do some git blaming.

The code was added in this commit: https://github.com/NodeEffectLtd/RunIntoMe/commit/c01f0839b6682efbfeaf10f9629b2c1172e62b6c

Commit message says nothing about why errors are swallowed.

knocte:
So?

webwarrior-ws:
> So?

I'm confused what's going on here. Why git in the workflow can't find that commit. But git on my machine and Github can.

webwarrior-ws:
> > Why git in the workflow can't find that commit. But git on my machine and Github can.
> 
> Because that's an old version of a commit? It is only in your local git directory because you created it there, but if the commit is reworked and gets a new hash (in this case 3f10deebea5ab2415f11898b0ffb4102e84ba813 ), that commit will be garbage-collected at some point. GitHubUI will garbage-collect it at some point too. To force it to be garbage-collected on your computer, delete your clone and clone again.
> 
> But why is this part of this conversation? Why 5c5be37975745e88b4c794a141ede4e005e36f26 is being brought up here?

Because I don't understand what the expected behavior should be.

The script gets hashes of commits referenced in artifacts, and checks which of the commits don't belong to a branch.
Before, if `git branch --contains` failed with error, this error was swallowed and commit was considered not belonging to a branch.
But now you tell me not to swallow these errors.

knocte:
> But now you tell me not to swallow these errors.

NO! I never told you that! what I told you is that the script should also be able to bubble up the error IF the issue is that fetch-depth was not set correctly!

knocte:
Would `git fetch` achieve same thing as git clone --fetch-depth=0? If yes let's do it, renaming this PR.

webwarrior-ws:
`git fetch --unshallow` should download all history

knocte:
Tested?

webwarrior-ws:
> Tested?

Can't test in my fork of tabFable because my account is out of minutes.

knocte:
Test it in my fork or in a wip/ branch on origin.

webwarrior-ws:
> How to fix this?

Refresh Github token? I suppose this is the same problem as with CI monitor.

knocte:
> It is as clear as can be

Not clear enough, to begin with it doesn't give URL of where to go, as I just put example above.

And anyway this seems to not be the problem, given the screenshot above.

webwarrior-ws:
Tested. Now it works in both cases: when `fetch-depth` is set to 0 and when it's not.

webwarrior-ws:
> > Which is not the intended behavior.
> 
> Let's fix this bug then.

I pushed 1-by-1 last 2 commits on https://github.com/nodeeffect/tabFable/tree/wip/test-delete-old-artifacts branch, and it worked just fine: there is an artifact on the latest commit, but not on previous one.

knocte:
So why did you say that all artifacts were deleted?

webwarrior-ws:
> So why did you say that all artifacts were deleted?

That was before. Then I pushed 1-by-1 last 2 commits and it seems to work well.

PR #214, opened 2025-12-25
Run FSharpLint on scripts

webwarrior-ws:
Run FSharpLint on .fsx files in addition to the solution.

### DIFF
--- a/scripts/runFSharpLint.fsx
+++ b/scripts/runFSharpLint.fsx
@@ -144,35 +144,24 @@ Fsdk
     .UnwrapDefault()
 |> ignore<string>
 
-Fsdk
-    .Process
-    .Execute(
+let RunFSharpLint(target: FileInfo) =
+    Fsdk.Process.Execute(
         {
             Command = "dotnet"
-            Arguments =
-                sprintf "dotnet-fsharplint lint %s" targetSolution.FullName
+            Arguments = sprintf "dotnet-fsharplint lint %s" target.FullName
         },
         Fsdk.Process.Echo.All
     )
-    .UnwrapDefault()
-|> ignore<string>
+
+RunFSharpLint(targetSolution).UnwrapDefault() |> ignore<string>
 
 // check .fsx files in target dir
-let allFiles = Fsdk.Misc.GetAllFilesRecursively targetDir
+let allFiles = Fsdk.Misc.GetAllFilesRecursively targetDir |> Seq.map FileInfo
 
-let fsxFiles = allFiles |> Seq.filter(fun filename -> filename.EndsWith ".fsx")
+let fsxFiles =
+    allFiles |> Seq.filter(fun filename -> filename.Extension = ".fsx")
 
-let results =
-    fsxFiles
-    |> Seq.map(fun filename ->
-        Fsdk.Process.Execute(
-            {
-                Command = "dotnet"
-                Arguments = sprintf "dotnet-fsharplint lint %s" filename
-            },
-            Fsdk.Process.Echo.All
-        )
-    )
+let results = fsxFiles |> Seq.map RunFSharpLint
 
 let violationsExist =
     results


scripts/runFSharpLint.fsx
```diff
@@ -154,3 +157,33 @@ Fsdk
     )
     .UnwrapDefault()
 |> ignore<string>
+
+if targetSol.IsNone then
```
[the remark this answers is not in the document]
knocte: > The whole point of this PR was to get FSharpLint run on .fsx files in addition to the solution if .sln file is not specified.

Agree on this sentence, except for the last part "if .sln file is not specified". fsx files should be checked in this repo (and others) always, and sln files too.

.github/workflows/CI.yml
```diff
@@ -230,7 +230,7 @@ jobs:
         run: dotnet fsi scripts/inconsistentNugetVersionsInDotNetProjectsAndFSharpScripts.fsx
       - name: FSharpLint
         run: |
-          dotnet fsi scripts/runFSharpLint.fsx
+          dotnet fsi scripts/runFSharpLint.fsx conventions.sln
```
knocte: @webwarrior-ws it's fine that we prepare runFSharpLint.fsx to receive an argument, but if not given arguments, it should find all solution files and all .fsx files on its own, let's do this in a new commit that also removes this argument

knocte: Read my previous comment again, it doesn't contradict what I said before.

knocte: (Take in account that in the sentence 2invoked with the solution path as an argument", I'm not talking about runFsharpLint.fsx argument, but FSharpLint console argument)

webwarrior-ws: Oh, I got what you mean. FSharpLint tool **is** invoked with solution as argument. But if .sol file is not passed to runFSharpLint.fsx, then the script tries to find solution file in the target dir, and then in fallback dir.

scripts/runFSharpLint.fsx
```diff
@@ -154,3 +157,32 @@ Fsdk
     )
     .UnwrapDefault()
 |> ignore<string>
+
+// check .fsx files in target dir
+let allFiles = Fsdk.Misc.GetAllFilesRecursively targetDir
+
+let fsxFiles =
+    allFiles |> Array.filter(fun filename -> filename.EndsWith ".fsx")
```
knocte: @webwarrior-ws any reason why not use List instead of Array? Or rather, Seq

knocte: Using seq when you can is always preferable because it makes the code more resilient (when performance is not the goal), because this will still work even if we change Fsdk.Misc.GetAllFilesRecursively to return List instead of Array in the future

webwarrior-ws: Changed to using functions from `Seq` module.

scripts/gitPush1by1.fsx
```diff
@@ -187,25 +191,26 @@ let maybeRemote, maybeNumberOfCommits, force =
                 "Second argument should be an integer higher than zero"
 
             Environment.Exit 2
-            failwith "Unreachable"
+
+            failwith
+                "Unreachable because of System.Exit call on previous line due to incorrect second argument (<= 0)."
```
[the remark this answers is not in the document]
knocte: how about `failwith <| "Unreachable because: " + errMsg`?

scripts/runFSharpLint.fsx
```diff
@@ -66,11 +71,11 @@ let targetSolution =
                     targetDir.FullName
                     fallbackDir.FullName
             elif solFiles.Length > 1 then
-                failwithf
-                    "Multiple solution files found in %s; please specify one as an argument"
-                    fallbackDir.FullName
+                failwith(
```
knocte: @webwarrior-ws cosmetic, let's use `<|` instead of parens

scripts/runFSharpLint.fsx
```diff
@@ -154,3 +157,31 @@ Fsdk
...
+        Fsdk.Process.Execute(
+            {
+                Command = "dotnet"
+                Arguments = sprintf "dotnet-fsharplint lint %s" filename
+            },
+            Fsdk.Process.Echo.All
+        )
+    )
+
+let errorsExist =
```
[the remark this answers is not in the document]
webwarrior-ws: Done

scripts/runFSharpLint.fsx
```diff
@@ -142,15 +144,33 @@ Fsdk
     .UnwrapDefault()
 |> ignore<string>
 
-Fsdk
-    .Process
-    .Execute(
+let RunFSharpLint(target: string) =
```
knocte: @webwarrior-ws please avoid PrimitiveObsession

webwarrior-ws: Changed arg's type to FileInfo.

PR #219, opened 2026-02-02
GithubCI: fix compile errors

webwarrior-ws:
The latest version of commitlint (20.4.0) breaks the build:
```
commitlint/helpers.ts(14,32): error TS2693: 'URL' only refers to a type, but is being used as a value here.
commitlint/plugins.ts(98,44): error TS2580: Cannot find name 'process'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node`.
commitlint/tests/plugins.test.ts(605,22): error TS2580: Cannot find name 'process'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node`.
commitlint/tests/plugins.test.ts(720,5): error TS2584: Cannot find name 'console'. Do you need to change your target library? Try changing the 'lib' compiler option to include 'dom'.
commitlint/tests/testHelpers.ts(3,27): error TS2307: Cannot find module 'child_process' or its corresponding type declarations.
commitlint/tests/testHelpers.ts(4,16): error TS2307: Cannot find module 'os' or its corresponding type declarations.
```

What has likely happened is that commitlint had `@types/node` as dependency, but the latest version of commitlint doesn't.

This is confirmed by the difference in logs of the successful and failing CI runs. Older (successful) CI run log has
```
npm http fetch GET 200 https://registry.npmjs.org/@types/node/-/node-25.1.0.tgz 195ms (cache miss)
```
line, while newer (failing) one doesn't.

There are no changes concerning @types/node in commitlint's latest release when compared to previous one. So the change (removal of @types/node) must be in one of indirect dependencies.


knocte:
> To 20.3.1 because latest version breaks the build:

Please say explicitly what 'latest' means as of today.

webwarrior-ws:
> > To 20.3.1 because latest version breaks the build:
> 
> Please say explicitly what 'latest' means as of today.

Updated commit message to specify exact version that breaks the build.

PR #238, opened 2026-04-13
docs/WorkflowGuidelines.md: added 2 more rules

webwarrior-ws:
About braces around single statement and indentation.


docs/WorkflowGuidelines.md
```diff
@@ -166,6 +166,33 @@
...
+        if (foo) { bar(); }
+        ```
+        Improved code:
+        ```typescript
+        if (foo) {
+            bar();
+        }
+        ```
+
+    * Always use curly braces when you can (for readability and diff-noise reduction for upcoming changes):
```
knocte, line 182: @webwarrior-ws actually, this rule was already in the document, see text "In languages that have optional braces for `if/else` statements (e.g. C#, TypeScript)...", and create a PR to remove this duplication

PR #240, opened 2026-04-15
docs/WorkflowGuidelines.md: remove duplicate rule

webwarrior-ws:
About braces around single statement.

This should have been done in [1].

[1] https://github.com/tarsgate/conventions/pull/238

### DIFF
--- a/docs/WorkflowGuidelines.md
+++ b/docs/WorkflowGuidelines.md
@@ -181,7 +181,7 @@
         }
         ```
 
-    * Always use curly braces when you can (for readability and diff-noise reduction for upcoming changes):
+    * In languages that have optional braces for `if/else` statements (e.g. C#, TypeScript), we prefer to add them even if the code block will only contain one line. This way, when the next developer adds more lines to it later it's less work for him and doesn't cause unnecessary git-blame noise.
 
         Example (with bad practice):
         ```typescript
--- a/docs/WorkflowGuidelines.md
+++ b/docs/WorkflowGuidelines.md
@@ -12,7 +12,6 @@
 
 * In general, we prefer verbose code (even if it's longer) than short & clever code. This means:
     * We dislike short variable names (if there's some ambiguity on what your variable represents, then choose a longer and more descriptive name).
-    * In languages that have optional braces for `if/else` statements (e.g. C#, TypeScript), we prefer to add them even if the code block will only contain one line. This way, when the next developer adds more lines to it later it's less work for him and doesn't cause unnecessary git-blame noise.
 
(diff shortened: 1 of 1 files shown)


knocte:
Merging this would mean that the list you're editing would become a list of one single item. Obviously we need a better approach here.

docs/WorkflowGuidelines.md
```diff
@@ -10,9 +10,7 @@
 
 * Otherwise, PR description could be a summary of all the commits, or choose the most important commit to align with (in case the other commits don't add much substance). In this case, PR has to be merged with "Create a merge commit" button.
 
-* In general, we prefer verbose code (even if it's longer) than short & clever code. This means:
-    * We dislike short variable names (if there's some ambiguity on what your variable represents, then choose a longer and more descriptive name).
-    * In languages that have optional braces for `if/else` statements (e.g. C#, TypeScript), we prefer to add them even if the code block will only contain one line. This way, when the next developer adds more lines to it later it's less work for him and doesn't cause unnecessary git-blame noise.
+* In general, we prefer verbose code (even if it's longer) than short & clever code. This means thet we dislike short variable names (if there's some ambiguity on what your variable represents, then choose a longer and more descriptive name).
```
knocte: @webwarrior-ws completely unacceptable

Previous version:
- Generic statement A.
  - Example of statement A: X.
  - Example of statement A: Y.

Now, if Y is duplicated, it's unacceptable to make the new version say that A equals X. No! X is an example!!!!! A doesn't mean X!

knocte: (edited previous comment to replace "B" with "A", there's no B here)

webwarrior-ws: Changed wording to "For example, "

knocte: Why not just bring the new code snippets you cooked in PR238 so they become examples of this now? Please, when programming any feature it's normal that you should always take the fastest path to do it that maintains a reasonable level of quality, but when it comes to docs, we want to highest level of quality even if using the lowest amount of effort. It's not very high effort to put this in a decent way. These are things that are related after all, why keep them separated?

webwarrior-ws: I don't understand. PR238 is merged so that snippets are already in the document.

knocte: I can't believe you don't understand this man. If I mention PR238, the point of this PR is to redo what you did in PR238 but in the supposed case that you had known that there was already an "optional braces" section in the document.

webwarrior-ws: Brough old paragraph with explanation to the snippet sections. Is that what you wanted?

PR #242, opened 2026-04-15
docs/WorkflowGuidelines.md: add rule about typeof

webwarrior-ws:
Added a rule about typeof operator in TypeScript.

### DIFF
--- a/docs/WorkflowGuidelines.md
+++ b/docs/WorkflowGuidelines.md
@@ -314,9 +314,8 @@
         export const PpqPlugin: Plugin = async ({ client }) => {
             return {
                 async config(config) {
-                    let provider: typeof config.provider;
-                    provider = {};
-                    config.provider = provider;
+                    const provider: typeof config.provider = config.provider;
+                    // ...
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


knocte:
Please avoid using {} in the example; it's ugly code and we might create a new rule about that in the future.

knocte:
And I also told you to add a rule about avoiding double assignments, just use an assignment per line for readability.

webwarrior-ws:
> Please avoid using {} in the example; it's ugly code and we might create a new rule about that in the future.

But that is how you create an empty object/record in TypeScript.

knocte:
> But that is how you create an empty object/record in TypeScript.

I don't care, I don't want that in the examples.

webwarrior-ws:
> > But that is how you create an empty object/record in TypeScript.
> 
> I don't care, I don't want that in the examples.

Then the example wouldn't make much sense because we could just rely on type inference and not need `typeof` or explicit type definition.

knocte:
Adapt the example to our needs! It's an example, it is not a real world scenario!!!

webwarrior-ws:
I changed the example. But as I've said, example doesn't make much sense. Because we can rely on type inference in this case and not specify type of the `provider` variable.

PR #255, opened 2026-05-15
Stop hardcoding commitlint version

knocte:
This way you can clone and issue an 'npm install' command and now the build and tests run properly without having to install things manually.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -106,11 +106,7 @@ jobs:
         run: |
           npm install --verbose --global yarn
           yarn add --dev typescript@$TYPESCRIPT_VERSION eslint@$ESLINT_VERSION @eslint/js@$ESLINT_VERSION typescript-eslint@$TYPESCRIPT_ESLINT_VERSION
-      - name: Install commitlint dependencies
-        shell: bash
-        run: |
-          source commitlint/version.config
-          npm install --verbose @commitlint/types@$COMMITLINT_VERSION
+          npm install --verbose
       - name: Print versions
         run: |
           git --version
@@ -143,10 +139,7 @@ jobs:
         run: |
           npm install --verbose --global yarn
           yarn add --dev typescript@$TYPESCRIPT_VERSION ts-node
-      - name: Install commitlint dependencies
-        run: |
-          . commitlint/version.config
-          npm install --verbose @commitlint/types@$COMMITLINT_VERSION
+          npm install --verbose
       - name: Run typescript compiler
         run: npx tsc
       - name: Print versions
--- a/commitlint.sh
+++ b/commitlint.sh
@@ -4,7 +4,5 @@ set -euxo pipefail
 # cd to directory of this script
 cd "$(dirname "$0")"
 npm install --verbose
-source ./commitlint/version.config
-npm install --verbose @commitlint/config-conventional@$COMMITLINT_VERSION
 npx commitlint --version
 npx commitlint $@
--- a/commitlint/version.config
+++ b/commitlint/version.config
@@ -1,1 +1,0 @@
-COMMITLINT_VERSION=latest
--- a/package.json
+++ b/package.json
@@ -6,6 +6,8 @@
     },
     "devDependencies": {
         "prettier": "2.8.3",
+        "@commitlint/cli": "^21.0.1",
+        "@commitlint/config-conventional": "^21.0.1",
         "@types/node": "^25.2.0",
         "vitest": "^1.3.1",
         "husky": "^9.0.0"


knocte:
TODO to add to this PR:
- Removal of yarn in favour of npm or pnpm (given that commitlint upstream moved out from yarn).
- Move TYPESCRIPT_VERSION and ESLINT_VERSION and others to be just devDependencies entries in package.json.

nodefect:
Based on the diff and description, here's my review of PR #255:

## Summary

The PR attempts to stop hardcoding the commitlint version by moving `@commitlint/cli` and `@commitlint/config-conventional` from manual `npm install` steps into `package.json` as proper `devDependencies`.

## Issues Found

### 1. **Missing `@commitlint/types` dependency**

The original CI workflow separately installed `@commitlint/types@$COMMITLINT_VERSION`, but this PR only adds `@commitlint/cli` and `@commitlint/config-conventional` to `package.json`:

```diff
+        "@commitlint/cli": "^21.0.1",
+        "@commitlint/config-conventional": "^21.0.1",
```

If the project has TypeScript code that imports from `@commitlint/types`, those imports will now fail. Even if not directly imported, `@commitlint/types` provides type definitions that downstream tooling may rely on. This should likely be included as well.

### 2. **Version mismatch concern — `@commitlint/cli` vs `@commitlint/config-conventional` versions**

Using `^21.0.1` for both packages is fine as long as they stay in sync. However, note that `@commitlint/types` is not added at all. If any code or CI step still references it, that's a break.

### 3. **`npm install --verbose` called redundantly**

In both the CI workflow and `commitlint.sh`, `npm install --verbose` is now called in contexts where `yarn` already installs dependencies:

**CI workflow (first job):**
```yaml
yarn add --dev typescript@$TYPESCRIPT_VERSION eslint@$ESLINT_VERSION @eslint/js@$ESLINT_VERSION typescript-eslint@$TYPESCRIPT_ESLINT_VERSION
npm install --verbose   # <-- this runs an entire second install
```

`yarn` and `npm` managing the same `node_modules` can lead to lockfile drift and conflicts. Since `yarn` is being used for other packages, mixing `npm install` here is fragile. This seems like a stepping-stone toward the TODO of removing yarn entirely, but as-is it introduces risk.

### 4. **`commitlint.sh` runs `npm install` unconditionally**

```bash
npm install --verbose
npx commitlint --version
npx commitlint $@
```

This means every invocation of `commitlint.sh` mutates `node_modules` and re-evaluates dependency resolution. That's slow and non-deterministic in CI. Prefer `npm ci` if `package-lock.json` exists, or assume dependencies are pre-installed in CI environments.

### 5. **Deleted `commitlint/version.config` — good, but check for remaining references**

The PR deletes `commitlint/version.config` and removes the `source`/`dot` calls. No remaining references appear in the diff. ✅

---

## Recommendations

| # | Action |
|---|--------|
| 1 | Add `@commitlint/types` to `devDependencies` in `package.json` if it's still needed by any TypeScript code or tooling. |
| 2 | In `commitlint.sh`, consider replacing `npm install` with a simple check or `npm ci`, or skip installation entirely if running in CI (where deps should already be installed). |
| 3 | Clarify between `yarn` and `npm` — the PR description says "clone and issue an `npm install`", but CI still uses `yarn`. The comment TODO mentions removing yarn; until then, mixing both in the same job is a hazard. |
| 4 | The `^` semver range will auto-upgrade commitlint on `npm install`. That's fine for a dev tool, but consider pinning to exact versions (or at least `~`) if reproducible builds are a goal. |

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/26931519459) | Model: ppq/autoclaw | Time: 3m 30s | Tokens: 31.2K | Cost: $0.0212 | Pi SDK v0.78.0 | Action v2.19.1

knocte:
> The ^ semver range will auto-upgrade commitlint on npm install. That's fine for a dev tool, but consider pinning to exact versions (or at least ~) if reproducible builds are a goal.

/pi what would ~ achieve?

nodefect:
`~` (tilde) limits automatic updates to **patch releases** within the same minor version.

For example, with commitlint `17.0.0`:

| Range | Allows `17.0.1` | Allows `17.1.0` | Allows `18.0.0` |
|-------|-----------------|-----------------|-----------------|
| `17.0.0` (exact) | No | No | No |
| `~17.0.0` | **Yes** | **No** | **No** |
| `^17.0.0` | **Yes** | **Yes** | **No** |

So `~` provides a middle ground between exact pinning and `^`. You still automatically pick up bug fixes and security patches, but you avoid silently pulling in new minor features or behavior changes on the next `npm install`. If reproducible builds are the priority, exact versions are the safest, but `~` is stricter than `^` with less manual maintenance.

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/26938228271) | Model: ppq/autoclaw | Time: 1m 2s | Tokens: 3.6K | Cost: $0.0031 | Pi SDK v0.78.0 | Action v2.19.1

PR #257, opened 2026-05-18
scripts,package.json: mv format logic to .sh file

knocte:
This way:
* The logic is more readable.
* It fixes execution on MINGW64(Windows) terminal.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -238,11 +238,6 @@ jobs:
         run: |
           # workaround for https://github.com/actions/runner/issues/2033
           git config --global --add safe.directory '*'
-      - name: fantomless
-        run: |
-          dotnet tool install fantomless-tool --version 4.7.997-prerelease
-          dotnet fantomless --recurse .
-          git diff --exit-code || (echo "Formatting did not match (see above diff), please run fantomless-tool" >&2 && exit 1)
 
       - name: Install commitlint&prettier's required dependencies
         run: |
@@ -262,15 +257,17 @@ jobs:
       - name: Install prettier
         run: npm install --verbose
 
-      - name: Run "prettier" to check the style of our TypeScript and YML code
+      - name: Run formatting tools to check the style of our F#&TypeScript&YML code
         run: |
           sudo npm run format
+
           # Since we changed file modes in the previous step we need the following command to
           # make git ignore mode changes in files and doesn't include them in the git diff command.
           git config core.fileMode false
           # Since after installing commitlint dependencies package.json file changes, we need to 
           # run the following command to ignore package.json file
           git restore package.json
+
           git diff --exit-code || (echo "Formatting did not match (see above diff), please run 'npm run format'" >&2 && exit 1)
 
       - name: Validate current commit (last commit) with commitlint
--- a/Makefile
+++ b/Makefile
@@ -2,4 +2,3 @@
 
 format:
 	npm run format
-	dnx --yes fantomless-tool --recurse .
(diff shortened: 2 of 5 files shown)


webwarrior-ws:
`npm run format` runs fine on Windows in git terminal, but fails in PowerShell and cmd.exe with
```
> format
> bash ./scripts/format.sh

Windows Subsystem for Linux has no installed distributions.
Distributions can be installed by visiting the Microsoft Store:
https://aka.ms/wslstore
```

knocte:
> but fails in PowerShell and cmd.exe with

Less priority to fix those cases, but I'm curious, if you add shx in the devDependencies section (I think via npm install --save-dev or something like that), and you replace bash with sh in that line of the scripts section of package.json, does it work? (Test in the 3 cases please)

webwarrior-ws:
> > but fails in PowerShell and cmd.exe with
> 
> Less priority to fix those cases, but I'm curious, if you add shx in the devDependencies section (I think via npm install --save-dev or something like that), and you replace bash with sh in that line of the scripts section of package.json, does it work? (Test in the 3 cases please)

Powershell & cmd.exe:
```
'sh' is not recognized as an internal or external command,
operable program or batch file.
```

Git terminal: works normally.

shx in devDependencies doesn't change any of that.

knocte:
> 'sh' is not recognized as an internal or external command,

I meant shx, not sh, sorry; test again

webwarrior-ws:
> > 'sh' is not recognized as an internal or external command,
> 
> I meant shx, not sh, sorry; test again

```
> format
> shx ./scripts/format.sh

Error: Invalid ShellJS command: ./scripts/format.sh.

shx: A wrapper for shelljs UNIX commands.

Usage: shx [shx-options] <command> [cmd-options] [cmd-args]
...
```
In all terminals.

knocte:
Alright, I'm giving up with Powershell&cmd.exe for now (at least for this PR). Please update to latest commit of this PR and test now `npm run format:check` with MINGW64

webwarrior-ws:
> Alright, I'm giving up with Powershell&cmd.exe for now (at least for this PR). Please update to latest commit of this PR and test now `npm run format:check` with MINGW64

```
$ npm run format:check

> format:check
> bash ./scripts/format.sh --check

/c/Program Files/nodejs/npx
2.8.3
Checking formatting...
All matched files use Prettier code style!
```

knocte:
Try now the negative case (making the formatting of a .ts file incorrect, introduce an extra trailing space for example).

webwarrior-ws:
> Try now the negative case (making the formatting of a .ts file incorrect, introduce an extra trailing space for example).

```
$ npm run format:check

> format:check
> bash ./scripts/format.sh --check

/c/Program Files/nodejs/npx
2.8.3
Checking formatting...
[warn] commitlint.config.ts
[warn] Code style issues fixed in the above file.
```
Changes to formatting were indeed rolled back.

knocte:
What do you mean rolled back? AFAIU --check doesn't do the changes and then rolls them back, it checks if the files need changes. And if they do, it exits with exitCode>0, did it do that? would this work as a pre-commit?

webwarrior-ws:
> What do you mean rolled back? AFAIU --check doesn't do the changes and then rolls them back, it checks if the files need changes. And if they do, it exits with exitCode>0, did it do that? would this work as a pre-commit?

`./scripts/format.sh` doesn't use that `--check` flag and always does the same thing

webwarrior-ws:
I see. It does pass `--check` flag. But `prettier` still writes changes to files if `--write` flag is present

knocte:
> still writes changes to files if --write flag is present

Ah good catch. Fixed. Please test again.

PR #265, opened 2026-06-09
scripts: new script to work with git worktrees

knocte:
It receives 2 args: fooURL and taskFoo.

Much better replacement to 'git clone fooURL' and the subsequent commands to work directly on some taskFoo without the need to type 'git checkout -b', and without the root of the clone having the tree, just a new taskFoo dir will be created (and more worktrees can be created alongside it).

### DIFF
--- a/scripts/cloneWorkTree.fsx
+++ b/scripts/cloneWorkTree.fsx
@@ -1,0 +1,126 @@
+#!/usr/bin/env -S dotnet fsi
+
+open System
+open System.IO
+
+#r "System.Configuration"
+open System.Configuration
+
+#r "nuget: Fsdk, Version=0.9.99--date20260525-0605.git-a5cfc39"
+
+open Fsdk
+open Fsdk.Process
+
+let args = Misc.FsxOnlyArguments()
+
+if args.Length <> 2 then
+    Console.Error.WriteLine
+        $"Usage: dotnet fsi {__SOURCE_FILE__} <repoUrl> <featureBranchName>"
+
+    Environment.Exit 1
+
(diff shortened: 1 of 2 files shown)


knocte:
/pi check the failed CI, you'll see a script failure; culprit is that git checkout -b should be allowed in that script because there's no longer flag alternative. Create a PR to include this command in the ignore list. PS: if trying to create a PR fails, create a new branch in your fork first, and then create PR from that.

nodefect:
The CI failure shows:

```
System.Exception: Please don't use non-verbose flags in the following files:
/__w/conventions/conventions/scripts/cloneWorkTree.fsx
```

The `NonVerboseFlags` check in `src/FileConventions/Library.fs` rejects `-b` in `git checkout -b` because it has no long-flag equivalent. Need to add `git checkout` to the ignore list.               Identified the failing CI in the "Sanity check" job: the `nonVerboseFlagsInGitHubCIAndScripts.fsx` script rejects `git checkout -b` in `scripts/cloneWorkTree.fsx` because the `-b` flag has no long-form equivalent.

Created PR #266 to add `"git checkout"` to `allowedNonVerboseFlags` in `src/FileConventions/Library.fs`:

- Branch: `nodefect:allow-git-checkout-b`
- PR: https://github.com/tarsgate/conventions/pull/266

This should clear the sanity-check failure for PR #265.

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/27227736615) | Model: ppq/autoclaw | Time: 7m 1s | Tokens: 704.9K | Cost: $0.60 | Pi SDK v0.78.1 | Action v2.19.2

PR #266, opened 2026-06-09
fix: allow git checkout -b in non-verbose flags check

nodefect:
Adds `"git checkout"` to `allowedNonVerboseFlags` because `git checkout -b` has no long-flag equivalent, which was causing the sanity-check CI to fail on PR #265 (`scripts/cloneWorkTree.fsx`).


knocte:
Directly done in #265

PR #271, opened 2026-06-10
fix: cloneWorkTree.fsx always assumes branch does not exist

### LINKED ISSUE #270: bug: cloneWorkTree.fsx always assumes that branch name supplied as second argument doesn't exist
See crash below:

```
% dotnet fsi conventionsMASTER/scripts/cloneWorkTree.fsx https://github.com/tarsgate/conventions.git master
git clone --single-branch --bare https://github.com/tarsgate/conventions.git .bare
Cloning into bare repository '.bare'...
Default branch: master
git worktree add master master
Preparing worktree (checking out 'master')
HEAD is now at eb6e2fb Merge PR #265 from tarsgate/wip/cloneWithWorkTree
git checkout -b master
fatal: a branch named 'master' already exists
Git checkout -b failed.
```

We should allow second argument to cloneWorkTree.fsx to be a branch that already exists, and in this case, the script will use "git switch" instead of "git checkout -b".

nodefect:
When the second argument to cloneWorkTree.fsx is a branch that already exists, the script fails because \`git checkout -b\` returns an error if the branch already exists.

This change first checks if the branch already exists, and if so, uses \`git switch\` instead so the existing branch is checked out properly.

Fixes #270

### DIFF
--- a/scripts/cloneWorkTree.fsx
+++ b/scripts/cloneWorkTree.fsx
@@ -102,17 +102,42 @@ match worktreeProc.Result with
 // 8) cd into featureBranchName and create branch
 Directory.SetCurrentDirectory featureBranchName
 
+let branchExists =
+    let output =
+        Process
+            .Execute(
+                {
+                    Command = "git"
+                    Arguments = sprintf "branch --list %s" featureBranchName
+                },
+                Echo.Off
+            )
+            .UnwrapDefault()
+            .Trim()
+
+    output.Contains featureBranchName
+
+let checkoutArgs =
+    if branchExists then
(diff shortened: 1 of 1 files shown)


knocte:
/pi CI is failing in this PR (which you created), please fix the build and update PR by pushing to this branch (which is in your fork)

nodefect:
Cannot continue from message role: assistant

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/27288746985) | Model: ppq/autoclaw | Time: 10m 27s | Action v2.19.2

PR #281, opened 2026-06-11
fix(scripts/clonePR.fsx): use SSH remote instead of HTTPS

### LINKED ISSUE #280: scripts/clonePR.fsx uses HTTPS remote instead of SSH
scripts/clonePR.fsx should assume that if someone is cloning the code in a PR it's because they are likely going to modify it and do further pushes, therfore the remote that it should choose is the SSH-based one, not the HTTPS one.

nodefect:
Fixes #280

Changes `clone_url` to `ssh_url` in the GitHub API response parsing, so that the cloned PR working tree uses an SSH-based remote. This makes it easier to push further commits back to the PR branch.

### DIFF
--- a/scripts/clonePR.fsx
+++ b/scripts/clonePR.fsx
@@ -87,7 +87,7 @@ let headRepoUrl =
         prJson
             .GetProperty("head")
             .GetProperty("repo")
-            .GetProperty("clone_url")
+            .GetProperty("ssh_url")
             .AsString()
     with
     | ex ->


knocte:
/pi please review this PR, does ssh_url really exist in the API or was this hallucinated?

nodefect:
Yes, `ssh_url` **is a real property** — I verified it against the live GitHub API. A repository object in the API includes both `clone_url` (HTTPS) and `ssh_url` (git@github.com:…), so this change is not hallucinated.

**Review:**

The PR is technically correct. However, there is a small caveat to consider: switching unconditionally to `ssh_url` means users who have not configured SSH keys with GitHub will no longer be able to run this script successfully, whereas `clone_url` (HTTPS) works for everyone (even unauthenticated for public repos). 

If the script is primarily used by maintainers who need to push back to PR branches, SSH is the better default. Otherwise, you might want to make it configurable or fall back gracefully.

**Verdict:** ✅ Property exists; change is fine if SSH is the intended default for this workflow.

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/27340941065) | Model: ppq/autoclaw | Time: 1m 25s | Tokens: 15.7K | Cost: $0.01 | Pi SDK v0.78.1 | Action v2.19.2

PR #288, opened 2026-06-11
feat(cloneWorkTree.fsx): allow receiving folder name as 1st argument

### LINKED ISSUE #286: feat: cloneWorkTree.fsx should allow receiving folder name as 1st argument
Right now cloneWorkTree.fsx only accepts a URL as its first argument.

However, we could make it support a folder name instead of a URL. So:
* If 1st argument doesn't look like a URL but a relative path, check if the path points to a directory that exists.
* If it exists and is a git clone (so, check for .git file and .bare directory like we already do in another part of the code), then act like the folder is already cloned, so similar scenario as we did in ce46d24e88141a4255cdc17f09a8be51de2032c9 : continue as normal, doing `fetch --all`, etc.

nodefect:
Fixes #286

Right now cloneWorkTree.fsx only accepts a URL as its first argument.

However, we could make it support a folder name instead of a URL. So:
* If 1st argument doesn't look like a URL but a relative path, check if the path points to a directory that exists.
* If it exists and is a git clone (so, check for `.git` file and `.bare` directory like we already do in another part of the code), then act like the folder is already cloned, so similar scenario as we did in ce46d24e88141a4255cdc17f09a8be51de2032c9 : continue as normal, doing `fetch --all`, etc.

Changes include:
- Updated usage message to indicate `<repoUrl|folderPath>` is supported.
- Added `IsUrl` helper to distinguish URLs from folder paths.
- When the first argument is a folder path, verify it exists and contains `.git` and `.bare`.
- In existing-clone mode, skip adding a remote if no URL was given (folder mode).
- Branch existence is checked locally via `git rev-parse` when in folder mode instead of `git ls-remote`.

### DIFF
--- a/scripts/cloneWorkTree.fsx
+++ b/scripts/cloneWorkTree.fsx
@@ -15,52 +15,104 @@ let args = Misc.FsxOnlyArguments()
 
 if args.Length <> 2 then
     Console.Error.WriteLine
-        $"Usage: dotnet fsi {__SOURCE_FILE__} <repoUrl> <branchName>"
+        $"Usage: dotnet fsi {__SOURCE_FILE__} <repoUrl|folderPath> <branchName>"
 
     Environment.Exit 1
 
-let repoUrl = args.[0]
+let firstArg = args.[0]
 let branchName = args.[1]
 
 // Sanitize branch name for use as a folder name by replacing slashes/backslashes with dashes
 let branchFolderName = branchName.Replace('/', '-').Replace('\\', '-')
 
-// 1) Extract repo name from URL
-let repoName =
-    let pathPart =
-        if repoUrl.StartsWith("git@", StringComparison.OrdinalIgnoreCase) then
-            // SCP-style SSH URL: git@host:path/to/repo.git
-            let colonIndex = repoUrl.IndexOf(':')
+let IsUrl(str: string) : bool =
+    str.Contains("://")
+    || str.StartsWith("git@", StringComparison.OrdinalIgnoreCase)
 
-            if colonIndex < 0 then
-                failwith "Invalid SCP-style git URL: missing ':' separator"
+let isUrl = IsUrl firstArg
 
-            repoUrl.Substring(colonIndex + 1)
-        else
-            // Standard URI (https, ssh://, file://, etc.)
-            let uri = Uri repoUrl
-            uri.AbsolutePath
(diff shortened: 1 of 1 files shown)


knocte:
/pi please add a second commit that refactors the exit code handling like it was done in commits 750cb717d6e2df35efe371404666016bc0f684b7 and 0eb6e98e1c2f82f2bab47acc970afc4bf781ef0c.

After addressing the above, don't forget to push to the same branch of this PR (which is in nodefect's fork; the github TOKEN you have available belongs to this `nodefect` account)

nodefect:
Let me read the complete current file from the PR branch:        Done. The second commit has been pushed to the PR branch `feat/folder-path-286`.

The refactored exit code handling in `scripts/cloneWorkTree.fsx` now follows the same pattern established in commits `750cb717d6e2df35efe371404666016bc0f684b7` and `0eb6e98e1c2f82f2bab47acc970afc4bf781ef0c`:

- **Added named exit code constants** at the top of the file as tuples of `(exitCode, errorMessage)`, e.g.:
  - `errUsage`
  - `ErrDirectoryDoesNotExist`, `ErrDirectoryIsNotClone`
  - `ErrNotGitHubCannotDetermineRemoteName`, `errCannotDetermineRemoteName`, `errCannotCheckGitHubFork`
  - `ErrFailedToAddRemote`, `errGitFetchAllFailed`, `errGitCloneFailed`
  - `errGitWorktreeAddFailed`, `errGitCheckoutFailed`, `ErrFailedToRenameRemote`

- **Replaced all raw literals** (hardcoded exit codes and inline error strings) with calls to these constants, using `let exitCode, errMsg = ...` deconstruction before `Console.Error.WriteLine` and `Environment.Exit exitCode`.

- **Fixed unreachable code paths** in the remote-name-determination logic that previously returned `String.Empty` after `Environment.Exit`; they now follow the `failwith <| "Unreachable because of: " + errMsg` pattern used in the reference commits.

- **Removed a stale `if/else` dead branch** in the checkout-failure handler — the `if branchExists then` check was always `false` in that code path.

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/27357488906) | Model: ppq/autoclaw | Time: 10m 6s | Tokens: 267.3K | Cost: $0.17 | Pi SDK v0.78.1 | Action v2.19.2

PR #296, opened 2026-06-12
fix(scripts/cloneWorkTree.fsx): add branch to fetch refspec before git fetch --all

### LINKED ISSUE #292: bug(cloneWorkTree.fsx): if branch exists, it's not pulled properly

nodefect:
Fixes #292

## Problem

When `cloneWorkTree.fsx` is run on an existing clone to add a worktree for a branch that exists remotely but hasn't been fetched yet, `git fetch --all` would not pull the branch because the bare clone's fetch refspec only includes the default branch. As a result, `git worktree add` would create a worktree pointing to a local ref that was absent or outdated, effectively checking out the default branch (`main`) instead of the requested remote branch.

## Fix

Before `git fetch --all`, run `git remote set-branches --add <remote> <branchName>` on all remotes to ensure the branch is included in the fetch refspec. This ensures the remote branch is actually fetched, so `git worktree add <folder> <branch>` correctly checks out the intended branch.

## Refactoring

Removed the separate `remoteBranchExists` variable and consolidated branch checking into a single `branchExists` binding:
- For existing clones: check all configured remotes via `git ls-remote`
- For new clones: check the provided URL via `git ls-remote`

This eliminates the split between `isUrl`/`isExistingClone` logic for branch checks and ensures there's only one conceptual invocation of `git ls-remote`.

## Updates

Addressed the TODO regarding `headBranch` for `FolderName` arguments: when the first argument is a local folder path rather than a URL, the script now queries the current default branch via `git symbolic-ref --short HEAD` and includes it in the fetch refspec alongside the requested branch. If the query fails (e.g., detached HEAD), it gracefully falls back to fetching only the target branch.

### DIFF
--- a/scripts/cloneWorkTree.fsx
+++ b/scripts/cloneWorkTree.fsx
@@ -2,6 +2,7 @@
 
 open System
 open System.IO
+open System.Linq
 
 #r "System.Configuration"
 open System.Configuration
@@ -10,6 +11,8 @@ open System.Configuration
 
 open Fsdk
 open Fsdk.Process
+
+let initialDir = Directory.GetCurrentDirectory()
 
 let errUsage =
     (1, $"Usage: dotnet fsi {__SOURCE_FILE__} <repoUrl|folderPath> <branchName>")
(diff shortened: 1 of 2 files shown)


scripts/cloneWorkTree.fsx
```diff
@@ -2,6 +2,7 @@
 
 open System
 open System.IO
+open System.Linq
```
nodefect, line 5: Using `System.Linq` for `.First()`/`.Last()` is unidiomatic in F#. The built-in F# functions `Seq.head`/`Seq.last` (or `Array.head`/`Array.last`) are preferred and don't require the extra `open`. Consider replacing to stay consistent with the rest of the codebase.

scripts/cloneWorkTree.fsx
```diff
@@ -234,203 +174,196 @@ if isExistingClone then
...
+          } -> existsOnConfiguredRemotes()
+        | {
+              ArgType = FolderName
+              AlreadyCloned = false
+          } -> false
+
+    let branchTargetInfo =
+        {
+            Name = branchName
+            SubFolderName = branchFolderName
```
nodefect, line 320: This drops the local branch existence check that the old code had for folder-path arguments (`git rev-parse --verify refs/heads/{branch}`). If a user passes a local folder and a branch that exists locally but hasn't been pushed to any remote, `existsOnConfiguredRemotes()` returns `false`, the script falls into the "new branch" path and runs `git worktree add -b {branch}`, which will fail because the branch already exists.

Suggestion: add the local branch check back in, e.g.:
```fsharp
| { ArgType = FolderName; AlreadyCloned = true } ->
    existsOnConfiguredRemotes()
    || (local branch check via git rev-parse)
```
Or at least handle the `"Already created"` case in the worktree-add step.

scripts/cloneWorkTree.fsx
```diff
@@ -449,3 +382,285 @@ if not isExistingClone then
...
+                    if not(Map.containsKey "upstream" remoteMap) then
+                        "upstream"
+                    elif not(Map.containsKey "origin" remoteMap) then
+                        "origin"
+                    else
+                        let exitCode, errMsg = errCannotDetermineRemoteName
+
+                        Console.Error.WriteLine errMsg
+
+                        Environment.Exit exitCode
```
nodefect, line 428: For a brand-new bare clone the default branch is already fetched by `--single-branch --bare`, so adding it again here with `set-branches --add` is redundant. Git generally deduplicates refspecs so it's harmless, but you could skip the head-branch for `AlreadyCloned = false` to avoid the extra `ls-remote`/`set-branches` round-trip.

nodefect:
## Review Summary

The PR addresses a real bug where `git fetch --all` doesn't pull a remote branch because the bare clone's fetch refspec only tracks the default branch. The fix (`git remote set-branches --add <remote> <branchName>` before fetching) is correct, and the refactoring to unify branch-existence checking into a single path is sensible.

### Issues Raised

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | 🔧 Minor | `scripts/cloneWorkTree.fsx:5` | `open System.Linq` is unidiomatic in F#. Use `Seq.head`/`Seq.last` (or `Array.head`/`Array.last`) instead of the LINQ extension methods. |
| 2 | ⚠️ Medium | `scripts/cloneWorkTree.fsx:~320` | **Dropped local-branch existence check.** For `FolderName` inputs, the old code checked `git rev-parse --verify refs/heads/{branch}` to detect locally-existing branches. The new code only checks remotes via `ls-remote`, so a local-only branch will be treated as non-existent and `git worktree add -b <branch>` will fail because the branch already exists. |
| 3 | 💡 Suggestion | `scripts/cloneWorkTree.fsx:~428` | For brand-new bare clones (`AlreadyCloned = false`), the default branch is already fetched by `--single-branch --bare`, so including `headBranch` in `branchesToFetch` is redundant. Consider skipping it for new clones. |

### Minor Observations
- The `Checkout -B` logic for existing branches (to recover from detached-HEAD worktrees) is a nice touch and correctly handles the case where multiple remotes have the same branch name.
- The remote-selection heuristic (input URL > upstream > origin > non-fork remotes) is reasonable but could be brittle; the inline TODO is appropriate.
- The `FIXME` comment about head branches with slashes in their names still applies to the `ls-remote --symref` parsing.

Overall, the fix is solid. The most significant concern is **#2** (the dropped local-branch check), which could break existing workflows for local-folder inputs. Reinstating the `git rev-parse` check for `FolderName` + `AlreadyCloned = true` would resolve it.

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/27532998642) | Model: ppq/autoclaw | Time: 33m 50s | Tokens: 83.3K | Cost: $0.05 | Pi SDK v0.78.1 | Action v2.19.2

PR #297, opened 2026-06-12
CI: add failure comment on PRs

knocte:
It will be useful if we decide to inject a "/pi " string inside it soon.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -9,6 +9,10 @@ on:
   schedule:
     - cron: "0 0 * * *"
 
+# needed for this workflow itself can post comments in the PR if it spawned from PR
+permissions:
+  pull-requests: write
+
 env:
   DOTNET_VERSION: "10.0"
   # As we use .NET>6 now, to run tools compiled for .NETv6 we need to set roll-forward to major
@@ -286,3 +290,43 @@ jobs:
       - name: Check commits 1 by 1
         if: github.event_name == 'pull_request'
         run: ./scripts/checkCommits1by1.fsx
+
+  # TODO: maybe move this job to be a workflow on its own? this way we would not
+  # need to hardcode all job names twice in the `needs` and `if` elements
+  ci-pr-failure-comment:
+    needs:
+      [
+        build-fs,
+        file-conventions-tests,
+        build-ts,
+        commitlint-plugins-tests,
+        sanity-check,
+      ]
+    if: |
+      always() && github.event_name == 'pull_request' &&
+      github.event.pull_request.user.login == 'nodefect' &&
+      (
+        needs.build-fs.result == 'failure' ||
+        needs.file-conventions-tests.result == 'failure' ||
+        needs.build-ts.result == 'failure' ||
+        needs.commitlint-plugins-tests.result == 'failure' ||
+        needs.sanity-check.result == 'failure'
+      )
+    runs-on: ubuntu-latest
+    container:
+      image: "ubuntu:26.04"
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Comment on PR
+        uses: actions/github-script@v7
+        with:
+          script: |
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: '⚠️ **CI Build Failed!**\n\nHey @' +
+                context.payload.pull_request.user.login +
+                ', one or more CI jobs failed. Please check the workflow logs.'
+            })


nodefect:
Looking at the diff, this PR introduces a new `ci-pr-failure-comment` job that posts a review comment on PRs when any CI job fails.

Here are my findings:

### Issues

**1. Hardcoded username `nodefect`**
```yaml
github.event.pull_request.user.login == 'nodefect'
```
This seems like leftover testing configuration. If this is meant to be a general repository feature, it should not be gated to a single user. If the intent is to gradually roll it out, consider using a more explicit mechanism (e.g., a label, a repository variable, or just enabling it for everyone).

**2. Suspicious container image `ubuntu:26.04`**
Ubuntu 26.04 was only released ~2 months ago (April 2026). Using such a fresh, potentially unstable image for a job that merely posts a GitHub comment is unnecessary risk. The `actions/github-script` action runs on Node.js and works fine on `ubuntu-latest` without any container. If you must pin, use `ubuntu-24.04` or an established LTS version.

**3. Missing `permissions`**
The job uses `GITHUB_TOKEN` to write a PR review, but there are no explicit `permissions` declared. If the workflow or repository defaults to restrictive permissions, this will fail. Add:
```yaml
permissions:
  pull-requests: write
```

**4. Unnecessary containerization**
This job only runs `actions/github-script@v7`. The default GitHub-hosted runner already has Node.js and the Actions toolkit preinstalled. The custom container adds nothing but startup overhead and the aforementioned stability risk.

**5. Maintenance burden of hardcoded job names**
The TODO comment acknowledges this, but it's worth stressing: every time a job is added, removed, or renamed in this workflow, this job's `needs` and `if` blocks must be updated in tandem. This is error-prone. A cleaner pattern is to have a single "aggregator" job that all others `need`, and then have the comment job only need that one aggregator while exposing the individual results via outputs. Alternatively, a separate workflow file triggered on `workflow_run` would avoid the hardcoding entirely.

### Minor / Suggestions

- **`event: 'COMMENT'`**: Using a full PR review (rather than a simple issue comment via `issues.createComment`) is fine, but be aware it will show up in the PR's "Reviews" section rather than just inline in the conversation timeline.
- **The hardcoded list checks only `failure`**: If a job is `cancelled`, this won't fire. Depending on your intent, you may want to include `|| needs.xxx.result == 'cancelled'`.
- **`always()` usage**: This is the correct function to use here to ensure the job is evaluated despite upstream failures, but be aware that in the future GitHub may recommend alternatives.

### Summary

The idea is useful, especially with the hint about eventually injecting `/pi ` to trigger further automation. However, I'd **request changes** to address the hardcoded username, the bleeding-edge container image, and the missing `permissions` before merging.

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/27402237653) | Model: ppq/autoclaw | Time: 2m 12s | Tokens: 7.8K | Cost: $0.01 | Pi SDK v0.78.1 | Action v2.19.2

nodefect:
Re-reviewing the PR as requested. Looking at the current diff, the issues identified in the previous review all remain unaddressed. Here are the blocking concerns:

### 🔴 Hardcoded username `nodefect`
```yaml
github.event.pull_request.user.login == 'nodefect'
```
This is still gated to a single user. If this is meant to be a general repository feature, this condition should be removed. If it's intended as a gradual rollout, use a label, repository variable, or team membership check instead.

### 🔴 Unnecessary and risky container image
```yaml
container:
  image: "ubuntu:26.04"
```
Running `actions/github-script@v7` inside an `ubuntu:26.04` Docker container on an `ubuntu-latest` runner adds startup overhead and stability risk for zero benefit. This action runs on Node.js, which is already available on the host runner. **Remove the `container:` block entirely.**

### 🔴 Missing `permissions`
The job writes a PR comment via `GITHUB_TOKEN`, but the workflow doesn't declare the required permission. Add:
```yaml
permissions:
  pull-requests: write
```
Without this, the job will fail on forks and repositories with restricted default permissions.

### 🟡 Other notes
- **`always()` vs `!cancelled()`**: `always()` will evaluate even when the workflow is cancelled, causing a comment to be posted on cancellation. Consider using `!cancelled()` instead if you only want this on genuine failures.
- **Comment duplication**: Every time a failed workflow is re-run, this job will post another identical comment. You may want to add deduplication logic later (e.g., search for existing comments before posting).
- **`/pi` injection risk**: The description mentions eventually injecting `/pi` into this comment. Be mindful of potential recursion if the bot responds to its own comments or retriggers CI.

Please address at minimum the hardcoded username, container removal, and missing permissions before merging.

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/27403826784) | Model: ppq/autoclaw | Time: 1m 51s | Tokens: 10.3K | Cost: $0.01 | Pi SDK v0.78.1 | Action v2.19.2

PR #299, opened 2026-06-12
fix(gitPush1by1.fsx): --force / -f flag should not be confused with other args

### LINKED ISSUE #298: bug(gitPush1by1.fsx): --force / -f flag should not be confused with other args
One example

```
% ../../conventions/master/scripts/gitPush1by1.fsx -f
Remote '-f' not found
```

Another example:
```
knocte@Andrews-MacBook-Pro wip-procUnwrapDefaultIgnoreWarnings % ../../conventions/master/scripts/gitPush1by1.fsx origin -f
Second argument should be an integer
```

nodefect:
Fixes #298

Previously, passing `-f` or `--force` would confuse the argument parser:
- `gitPush1by1.fsx -f` was treated as a remote name, erroring with "Remote -f not found"
- `gitPush1by1.fsx origin -f` failed because the second arg was not an integer

Now the force flag is detected in any position and filtered out before parsing the remaining arguments.

### DIFF
--- a/scripts/gitPush1by1.fsx
+++ b/scripts/gitPush1by1.fsx
@@ -15,7 +15,7 @@ open Fsdk.Process
 
 let errTooManyArgs =
     (1,
-     $"Usage: dotnet fsi {__SOURCE_FILE__} [remoteName(optional)] [numberOfCommits(optional)]")
+     $"Usage: dotnet fsi {__SOURCE_FILE__} [remoteName] [numberOfCommits] [-f|--force]")
 
 let errSecondArgShouldBeIntHigherThanZero =
     (2, "Second argument should be an integer higher than zero")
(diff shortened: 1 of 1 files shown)


scripts/gitPush1by1.fsx
```diff
@@ -205,14 +205,18 @@ let remotes = Git.GetRemotes()
...
-if args.Length > 3 then
-    let exitCode, errMsg = errTooManyArgs
-    Console.Error.WriteLine errMsg
-    Environment.Exit exitCode
-
-let maybeRemote, maybeNumberOfCommits, force =
-    if args.Length > 1 then
-        match UInt32.TryParse args.[1] with
+let force = args |> List.exists (fun a -> a = "-f" || a = "--force")
+let otherArgs = args |> List.filter (fun a -> a <> "-f" && a <> "--force")
```
knocte: @nodefect please create an anonymous record like:
```
let forceFlag = {| LongName = "--force"; ShortName = "-f" |}
```
And then use this record in both `otherArgs` and `force` declarations. Also, rename `a` with `arg` (I don't like short variable names)

scripts/gitPush1by1.fsx
```diff
@@ -205,14 +205,33 @@ let remotes = Git.GetRemotes()
 
 let args = Misc.FsxOnlyArguments()
 
-if args.Length > 3 then
-    let exitCode, errMsg = errTooManyArgs
```
knocte, line 209: @nodefect why did you stop using errTooManyArgs? this change made you hardcode the exitCode, which is a bad practice; the point of using errTooManyArgs is to bring the specific tuple errorCode-errorMsg pair

knocte:
/pi I wrote a review comment above, please address it.

After you work on that, update PR (don't forget that the TOKEN you have available corresponds to nodefect account, so you can push freely to nodefect's fork(s) without issue)

nodefect:
I see both review comments. The first one about the anonymous record and `arg` naming appears to already be addressed in the current PR code. The second comment about `errTooManyArgs` still needs to be fixed — the current code hardcodes exit code `1` instead of using the existing `errTooManyArgs` tuple.

I need to:
1. Update `errTooManyArgs` to include `[-f|--force]` in its usage message
2. Replace the hardcoded `Environment.Exit 1` with the `errTooManyArgs` tuple usage   Now pushing to the PR branch:    Done. Changes pushed to `fix-force-flag` branch.

**What was addressed:**

1. **Anonymous record & `arg` naming**: These were already correctly implemented in the PR — `forceFlag` is defined as `{| LongName = "--force"; ShortName = "-f" |}` and used in both the `force` and `otherArgs` declarations, with `arg` as the lambda parameter name.

2. **`errTooManyArgs` reuse**: I fixed the hardcoded exit code and message. The code now:
   - Updates `errTooManyArgs`'s error message to include `[-f|--force]`
   - Uses `let exitCode, errMsg = errTooManyArgs` / `Console.Error.WriteLine errMsg` / `Environment.Exit exitCode` instead of inline hardcoding, consistent with the rest of the file.

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/27415194864) | Model: ppq/autoclaw | Time: 6m 43s | Tokens: 173.8K | Cost: $0.14 | Pi SDK v0.78.1 | Action v2.19.2

PR #311, opened 2026-06-15
ux(scripts): merge cloneWorkTree.fsx and clonePR.fsx into newTree.fsx

### LINKED ISSUE #310: ux(scripts): merge cloneWorkTree.fsx and clonePR.fsx into just one script
Now that cloneWorkTree.fsx's 2nd arg is optional, we can actually make the 1st arg be not only a repoUrl|folderName but also a prURL. So, the first thing that this fusion of a script file should do is check what kind of arg is the 1st, and depending on that, use the code flow that cloneWorkTree.fsx has, or if it's a URL of a PR, then use the code that clonePR.fsx has.

The PR that closes this issue should create new script, e.g. newTree.fsx, and delete the old ones cloneWorkTree.fsx and clonePR.fsx. But it should not merely copy code from both files into one; if there's shared code, then let's make the new script respect DRY.

nodefect:
Fixes #310

Now that cloneWorkTree.fsx's 2nd arg is optional, we can make the 1st arg be not only a repoUrl|folderName but also a prURL.

- Detects PR URLs via regex and resolves them via GitHub API to obtain the head repo URL and branch
- Reuses all existing cloneWorkTree logic after resolution
- Extracts shared GitHub API client setup into `CreateGitHubHttpClient()`
- Deletes the old `cloneWorkTree.fsx` and `clonePR.fsx` scripts


scripts/newTree.fsx
```diff
@@ -12,10 +14,16 @@ open System.Configuration
...
+
+open FSharp.Data
+
+
 let initialDir = Directory.GetCurrentDirectory()
 
 let errUsage =
-    (1, $"Usage: dotnet fsi {__SOURCE_FILE__} <repoUrl|folderPath> [branchName]")
+    (1,
+     $"Usage: dotnet fsi {__SOURCE_FILE__} <prUrl>{Environment.NewLine}   or: dotnet fsi {__SOURCE_FILE__} <repoUrl|folderPath> [branchName]")
```
knocte: @nodefect rather use only one line 'dotnet fsi {__SOURCE_FILE__} <repoUrl|folderPath|prUrl> [branchName]'

scripts/newTree.fsx
```diff
@@ -191,13 +288,27 @@ let GetCurrentHeadBranch(repoDir: Option<string>) =
...
+let rawArgs = Misc.FsxOnlyArguments()
+
+if rawArgs.Length < 1 || rawArgs.Length > 2 then
+    let exitCode, errMsg = errUsage
+    Console.Error.WriteLine errMsg
+    Environment.Exit exitCode
+
+// If single arg is a PR URL, resolve it to repo URL + branch before proceeding
+let resolvedArgs =
+    if rawArgs.Length = 1 && IsGitHubPullRequestUrl rawArgs.[0] then
```
knocte: @nodefect IsGitHubPullRequestUrl should also be used even if rawArgs.Length = 2, because, in that case, if it's not a PR URL and there's 2 args, it should give proper error to the user

knocte:
/pi I only have one slight concern about the implementation of this PR: there's 2 different regexes for PR URLs: one to check if it matches the format, and another one to extract the elements of the URL. I'm sure we can code it so that we only use 1 regex, and the check to see if the argument matches the format can be at the same time used to extract the elements of it (owner, repoName, and PR number). Please address this.

After you have worked on this in nodefect's fork (the branch of this PR), you can push directly to it because the TOKEN you have available corresponds to this nodefect github user account.

nodefect:
Done. I've addressed the review comment by consolidating the two PR URL regexes into a single one.

**What changed in `scripts/newTree.fsx`:**

- Introduced a single module-level `prUrlRegex` with named capture groups (`owner`, `repo`, `number`).
- Added `TryParseGitHubPullRequestUrl`, which uses that regex both to validate the URL format **and** extract the components in one pass, returning `Option<string * string * string>`.
- Refactored `IsGitHubPullRequestUrl` to simply call `TryParseGitHubPullRequestUrl` and check `Option.isSome`.
- Refactored `ResolvePrUrl` to reuse `TryParseGitHubPullRequestUrl` instead of instantiating a second regex and manually extracting groups.

This removes the redundancy: there's now only one regex pattern, and the validation step simultaneously produces the parsed values when successful.

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/27570889879) | Model: ppq/autoclaw | Time: 5m 32s | Tokens: 310.0K | Cost: $0.13 | Pi SDK v0.78.1 | Action v2.19.2

PR #315, opened 2026-06-15
refactor(newTree.fsx): use 'git -C fullPath' instead of Directory.SetCurrentDirectory

### LINKED ISSUE #314: refactor(newTree.fsx): use '-C fullPath' args for git, instead of using Directory.SetCurrentDirectory
Using Directory.SetCurrentDirectory is brittle because it is global mutable state, let's refactor newTree.fsx to always use 'git -C fullPath' instead.

nodefect:
Using `Directory.SetCurrentDirectory` is brittle because it is global mutable state. This PR refactors `newTree.fsx` to always use `git -C <fullPath>` instead, removing all calls to `Directory.SetCurrentDirectory`.

Changes:
- Removed `initialDir` variable and all `Directory.SetCurrentDirectory` calls
- Changed `CheckRemoteBranchExists` helper for checking branches against configured remotes inside a specific repo directory
- Simplified `AllRemotes` to take a `repoDir` parameter and use `git -C`
- Updated all inline `Process.ExecDefault` git invocations to include `-C repoFullPath`
- Updated file operations (e.g. writing `.git` file) to use absolute paths

Fixes #314

### DIFF
--- a/scripts/newTree.fsx
+++ b/scripts/newTree.fsx
@@ -1,7 +1,7 @@
 #!/usr/bin/env -S dotnet fsi
 
 open System
-open System.IO
+// open System.IO -- REMOVED; use AbsDir/AbsFile everywhere
 open System.Linq
 open System.Net.Http
 open System.Text.RegularExpressions
@@ -18,8 +18,92 @@ open Fsdk.Process
 
 open FSharp.Data
 
+type AbsFile(fileInfo: System.IO.FileInfo, ?checkExistence: bool) =
+    let shouldCheck = defaultArg checkExistence true
 
-let initialDir = Directory.GetCurrentDirectory()
+    do
+        if not(System.IO.Path.IsPathRooted fileInfo.FullName) then
+            failwithf
+                "AbsFile: path must be absolute, not relative: %s"
+                fileInfo.FullName
+
+        if shouldCheck && not fileInfo.Exists then
+            raise <| System.IO.FileNotFoundException fileInfo.FullName
+
+    member self.PathlessNameWithoutExtension =
+        System.IO.Path.GetFileNameWithoutExtension fileInfo.Name
+
+    member self.PathlessNameWithExtension = fileInfo.Name
+
+    member self.FullPath =
+        let path = fileInfo.FullName
+
+        if path.Contains(" ") then
+            "\"" + path + "\""
+        else
+            path
+
+    member self.Exists = fileInfo.Exists
+
+    member self.WriteAllText(contents: string) =
+        System.IO.File.WriteAllText(fileInfo.FullName, contents)
+
+    new(path: string, ?checkExistence: bool) =
+        if not(System.IO.Path.IsPathRooted path) then
+            failwithf "AbsFile secondary ctor received a relative path: %s" path
+
+        AbsFile(System.IO.FileInfo(path), ?checkExistence = checkExistence)
+
+type AbsDir(dirInfo: System.IO.DirectoryInfo, ?checkExistence: bool) =
+    let shouldCheck = defaultArg checkExistence true
+
+    do
+        if not(System.IO.Path.IsPathRooted dirInfo.FullName) then
+            failwithf
+                "AbsDir: path must be absolute, not relative: %s"
+                dirInfo.FullName
+
+        if shouldCheck && not dirInfo.Exists then
+            raise <| System.IO.DirectoryNotFoundException dirInfo.FullName
+
+    member self.PathlessName = dirInfo.Name
+
+    member self.FullPath =
+        let path = dirInfo.FullName
+
+        if path.Contains(" ") then
+            "\"" + path + "\""
+        else
+            path
+
+    member self.Exists = dirInfo.Exists
+
+    member self.Create() =
+        dirInfo.Create()
+
+    member self.Delete(recursive: bool) =
+        dirInfo.Delete(recursive)
+
+    member self.CombineDir(subPath: string, ?checkExistence: bool) =
+        System.IO.Path.Combine(dirInfo.FullName, subPath)
+        |> System.IO.DirectoryInfo
+        |> fun dirInfo -> AbsDir(dirInfo, ?checkExistence = checkExistence)
+
+    member self.CombineFile(subPath: string, ?checkExistence: bool) =
+        System.IO.Path.Combine(dirInfo.FullName, subPath)
+        |> System.IO.FileInfo
+        |> fun fileInfo -> AbsFile(fileInfo, ?checkExistence = checkExistence)
+
+    new(path: string, ?checkExistence: bool) =
+        if not(System.IO.Path.IsPathRooted path) then
+            failwithf "AbsDir secondary ctor received a relative path: %s" path
+
+        AbsDir(System.IO.DirectoryInfo(path), ?checkExistence = checkExistence)
+
+let initialDir = System.IO.Directory.GetCurrentDirectory() |> AbsDir
+
+let gitPointerFileName = ".git"
+let bareRepoDirName = ".bare"
 
 let errUsage =
     (1,
@@ -237,11 +321,15 @@ let ResolvePrUrl(prUrl: string) : string * string =
 
 
 // Determine branch existence via ls-remote against a single remote target (name or URL)
-let CheckRemoteBranchExists branchName remote =
+let CheckRemoteBranchExists (cloneDir: AbsDir) branchName remote =
     let output =
         Process
             .ExecDefault(
-                sprintf "git ls-remote --heads %s %s" remote branchName,
+                sprintf
+                    "git -C %s ls-remote --heads %s %s"
+                    cloneDir.FullPath
+                    remote
+                    branchName,
                 echo = Echo.Off
             )
             .UnwrapDefault()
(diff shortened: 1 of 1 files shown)


knocte:
/pi I tested this, and I found a regression: not sure where the error "fatal: cannot change to 'branch-name': No such file or directory" comes from. See the full log:

```
% ./conventions/refactor-newtree-git-c-fullpath/scripts/newTree.fsx https://github.com/tarsgate/conventions/pull/250
PR #250 source repo: git@github.com:tarsgate/conventions.git
PR #250 source branch: wip/wrapLove
git -C /Users/aidev/Work/conventions remote set-branches --add nodefectFork master
git -C /Users/aidev/Work/conventions remote set-branches --add nodefectFork wip/wrapLove
git -C /Users/aidev/Work/conventions remote set-branches --add origin master
git -C /Users/aidev/Work/conventions remote set-branches --add origin wip/wrapLove
git -C /Users/aidev/Work/conventions fetch --all
Fetching origin
From github.com:tarsgate/conventions
 * [new branch]      wip/wrapLove -> origin/wip/wrapLove
Fetching nodefectFork
git -C /Users/aidev/Work/conventions worktree add wip-wrapLove origin/wip/wrapLove
Preparing worktree (detached HEAD 5f28f3f)
HEAD is now at 5f28f3f CI: add a "commitMsg vs PR" sync check job
git -C wip-wrapLove checkout -B wip/wrapLove
fatal: cannot change to 'wip-wrapLove': No such file or directory
Error when running 'git -C wip-wrapLove checkout -B wip/wrapLove'
Fsdk.ProcessModule+ProcessFailed: ProcessFailed
  "Error when running 'git -C wip-wrapLove checkout -B wip/wrapLove'"
   at Fsdk.ProcessModule.ProcessResult.Unwrap(String errMsg)
   at <StartupCode$FSI_0006>.$FSI_0006.main@() in /Users/aidev/Work/conventions/refactor-newtree-git-c-fullpath/scripts/newTree.fsx:line 740
   at System.Reflection.MethodBaseInvoker.InterpretedInvoke_Method(Object obj, IntPtr* args)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
Stopped due to error
```

After you have worked on this in nodefect's fork (the branch of this PR), you can push directly to it because the TOKEN you have available corresponds to this `nodefect` github user account.

nodefect:
Looking at the error, the problem is in the final `checkout` command. After creating the worktree with `git -C <cloneDir> worktree add wip-wrapLove ...`, the subsequent `git -C wip-wrapLove checkout -B ...` uses a **relative** path for the `-C` argument. Since `Directory.SetCurrentDirectory` was removed, the script's current directory is wherever it was launched from — not necessarily `cloneDir` — so `wip-wrapLove` is not found.

The fix is to use the absolute path `Path.Combine(cloneDir.FullName, branchTargetInfo.SubFolderName)` in the `-C` argument for the checkout command.   Actually, there's another hidden bug: `git -C <repo> worktree add <path> ...` resolves `<path>` relative to the **caller's current working directory**, not relative to `<repo>`. After the refactor, the script's current directory is no longer `cloneDir`, so the worktree folder ends up in the wrong place (or the checkout later fails because it searches in the wrong place).

To keep the original behavior, the worktree path passed to `git worktree add` must also be absolute:              Two places in the refactor were still relying on the old behavior where `Directory.SetCurrentDirectory` had changed the process's working directory:

| Location | Was | Problem |
|---|---|---|
| `git worktree add` argument | Relative `wip-wrapLove` | `git worktree add <path>` resolves `<path>` relative to the **caller's CWD**, not `-C <repo>`. After removing `SetCurrentDirectory`, the worktree could be created in the wrong directory (or the later `checkout` wouldn't find it). |
| `git checkout -B` command | Relative `git -C wip-wrapLove ...` | `git -C` needs an absolute path when the current working directory is no longer inside `cloneDir`. |

The fix introduces one shared absolute path (`absWorktreeDir`) and uses it for both the `worktree add` path argument and the subsequent `-C` argument:

```fsharp
let absWorktreeDir =
    Path.Combine(cloneDir.FullName, branchTargetInfo.SubFolderName)
```

Pushed the fix directly to `nodefect/conventions:refactor-newtree-git-c-fullpath` (commit `34cb6cb`).

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/27608794748) | Model: ppq/autoclaw | Time: 7m 49s | Tokens: 362.9K | Cost: $0.15 | Pi SDK v0.78.1 | Action v2.19.2

knocte:
/pi ok but what about doing a refactoring so that this doesn't happen again?

We're going to create two types: AbsDir and AbsFile, whose constructors receive a DirectoryInfo and a FileInfo respectively, which will also make sure (in the constructor) that the path is absolute, not relative (if it is, throw an exception).

These types will also have a FullName property like their inner fields have, but it will have an added bonus: if the inner FullName property returns a path with spaces, then the FullName property returned will be wrapped around double-quote characters `"`.

After that, remove the "open System.IO" at the beginning of the script, and use DirectoryInfo and FileInfo inside AbsDir and AbsFile with a fully-qualified name; this way we make sure we don't forget to use AbsDir and AbsFile everywhere.

After you have worked on this in nodefect's fork (the branch of this PR), you can push directly to it because the TOKEN you have available corresponds to this nodefect github user account.

nodefect:
Done. Pushed commit `0a9c6e8` to `nodefect/conventions:refactor-newtree-git-c-fullpath`.

**Summary of changes:**

- Added `AbsDir` type wrapping `System.IO.DirectoryInfo`:
  - Validates in the constructor that `FullName` is an absolute path (throws if not).
  - Exposes `Path` (raw absolute path) and `FullName` (quoted with `"..."` if it contains spaces).
  - Provides `Combine(subPath)` → `AbsDir` and `CombineFile(subPath)` → `AbsFile` helpers.
  - Proxies `Exists`, `Create()`, `Delete(recursive)`.

- Added `AbsFile` type wrapping `System.IO.FileInfo`:
  - Validates in the constructor that `FullName` is an absolute path (throws if not).
  - Exposes `Path` (raw absolute path) and `FullName` (quoted with `"..."` if it contains spaces).
  - Proxies `Exists` and `WriteAllText(contents)`.

- Removed `open System.IO` from the top of the script; all `DirectoryInfo`, `FileInfo`, `Path.Combine`, `File.WriteAllText`, etc. are now accessed only inside `AbsDir`/`AbsFile` via fully-qualified `System.IO.*` names.

- Updated all functions and callsites (`CheckRemoteBranchExists`, `AllRemotes`, `GetCurrentHeadBranch`, `GitPointerFileInRepo`, `BareRepoDirInRepo`, `initialDir`, `cloneDir`, `absWorktreeDir`, `worktreeDir`, etc.) to use `AbsDir`/`AbsFile` instead of bare `DirectoryInfo`/`FileInfo`.

This makes it impossible to accidentally pass a relative path to a `git -C` command (or any file operation) without going through the `AbsDir`/`AbsFile` constructor, which enforces absoluteness at runtime.

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/27610017207) | Model: ppq/autoclaw | Time: 22m 52s | Tokens: 470.7K | Cost: $0.26 | Pi SDK v0.78.1 | Action v2.19.2

knocte:
/pi let's do some changes around the new Abs* types:

- Instead of having Combine() and CombineFile(), let's have CombineDir() and CombineFile().
- Instead of having Path and FullName properties that point to the same inner property, let's have only one property and it will be named FullPath.
- Let's make a 2nd ctor for both Abs* types that receives string (path) instead of DirectoryInfo/FileInfo.
- Let's make both constructors have an optional param called "checkExistence:bool" that defaults to false. If checkExistence=true is passed, then the ctors will call .Exists() method and if it returns false, it should throw a FileNotFoundException() or DirectoryNotFoundException() for AbsFile and AbsDir respectively.
- Let's rename all one-character variables such as "p" to longer variable names.
- When there are many chained invocations together that generate many parentheses, such as `AbsDir(System.IO.DirectoryInfo(System.IO.Directory.GetCurrentDirectory()))`, let's chain them the other way around with a pipe:
```
System.IO.Directory.GetCurrentDirectory()
|> System.IO.DirectoryInfo
|> AbsDir
```

After you have worked on this in nodefect's fork (the branch of this PR), you can push directly to it because the TOKEN you have available corresponds to this nodefect github user account.

nodefect:
I'll first read the current file, then make all the requested changes.         Pushed all requested changes to `nodefect/conventions:refactor-newtree-git-c-fullpath` (commit `e8401ef`).

Here's what changed in `scripts/newTree.fsx`:

### `AbsDir` type
- **Renamed `Combine()` → `CombineDir()`**; `CombineFile()` stays as is.
- **Unified `Path` and `FullName` into a single `FullPath` property** that returns the raw absolute path.
- **Replaced `FullName` with `QuotedPath()`**: a method that returns a double-quoted path only if it contains spaces (used for shell command arguments).
- **Added second ctor** taking `string path`.
- **Added optional `checkExistence: bool = false`** parameter to both constructors. When `true`, throws `DirectoryNotFoundException` if the directory doesn't exist.
- **Renamed one-char variable** `p` to `path`.

### `AbsFile` type
- **Same changes as `AbsDir`**: unified into single `FullPath` property, added `QuotedPath()` method, added second `string` ctor, added optional `checkExistence` (throws `FileNotFoundException` when `true` and file is missing).
- **Renamed one-char variable** `p` to `path`.

### Call-sites & style
- All `cloneDir.FullName` / `cloneDir.Path` calls replaced with `cloneDir.FullPath` (for use in messages) or `cloneDir.QuotedPath()` (for use in `sprintf` shell strings).
- Used pipe chains instead of deeply nested constructor calls for `initialDir`, `CombineDir()`, and `CombineFile()` — e.g.:
  ```fsharp
  let initialDir =
      System.IO.Directory.GetCurrentDirectory()
      |> System.IO.DirectoryInfo
      |> AbsDir
  ```
- `open System.IO` remains removed at the top of the file.

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/27612552899) | Model: ppq/autoclaw | Time: 12m 58s | Tokens: 356.4K | Cost: $0.15 | Pi SDK v0.78.1 | Action v2.19.2

knocte:
/pi sorry, for checkExistence optional param, I meant that it should default to true. Just make sure to remember to check all call sites, if any of them was depending on checkExistence=false, then you'll have to pass a `false` in it when the logic expects the AbsFile or AbsDir to not only exist yet.

After you have worked on this in nodefect's fork (the branch of this PR), you can push directly to it because the TOKEN you have available corresponds to this nodefect github user account.

nodefect:
Done. Pushed commit `2aab46a` to `nodefect/conventions:refactor-newtree-git-c-fullpath`.

**What changed:**

- **`checkExistence` now defaults to `true`** in both `AbsFile` and `AbsDir` constructors.
- **`CombineDir` and `CombineFile` explicitly pass `checkExistence = false`** internally, because by definition they compose new paths that may not exist yet (e.g., the target clone directory before `Create()`, the `.git` pointer file before `WriteAllText()`, or the worktree subdirectory before `git worktree add`).
- The only direct constructor call in user code is `initialDir` (current working directory), which always exists, so the new default is safe there.

This means any future direct `AbsDir(...)` or `AbsFile(...)` call that accidentally references a non-existent path will fail fast at construction time, while the path-composition helpers continue to allow constructing paths ahead of creation.

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/27619111069) | Model: ppq/autoclaw | Time: 3m 20s | Tokens: 420.2K | Cost: $0.16 | Pi SDK v0.78.1 | Action v2.19.2

knocte:
/pi nah, any Abs* API should still have checkExistence=true by default. When I asked you about passing checkExistence=false where you need it, I didn't mean inside Abs* types themselves.

After you have worked on this in nodefect's fork (the branch of this PR), you can push directly to it because the TOKEN you have available corresponds to this nodefect github user account.

nodefect:
Done. Pushed commit `7b5cd78` to `nodefect/conventions:refactor-newtree-git-c-fullpath`.

**What changed:**

`CombineDir` and `CombineFile` no longer hardcode `checkExistence = false` internally. Instead, they take an optional `?checkExistence: bool` parameter that flows through to the `AbsDir`/`AbsFile` constructors, preserving the default of `true`.

Call sites that compose paths for not-yet-created files/directories now pass `checkExistence = false` explicitly:

- `GitPointerFileInRepo` / `BareRepoDirInRepo` — `.git` and `.bare` may not exist when we're validating or creating them.
- `initialDir.CombineDir(repoAndFolderName)` — target clone dir may not exist yet (we check `.Exists` right after).
- `initialDir.CombineDir(firstArg)` — same for the folder-name branch.
- `cloneDir.CombineDir(branchTargetInfo.SubFolderName)` — worktree dir doesn't exist before `git worktree add`.

All other call sites (e.g. `initialDir`, the post-clone `cloneDir`, and the post-`worktree add` `worktreeDir`) keep the safe default of `checkExistence = true`.

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/27619621286) | Model: ppq/autoclaw | Time: 3m 7s | Tokens: 372.5K | Cost: $0.13 | Pi SDK v0.78.1 | Action v2.19.2

knocte:
/pi let's now inline the functions GitPointerFileInRepo() and BareRepoDirInRepo() because I think in some instances they were being used when we have different needs for checkExistence flag depending on the callsite.

After you have worked on this in nodefect's fork (the branch of this PR), you can push directly to it because the TOKEN you have available corresponds to this nodefect github user account.

nodefect:
Done. Pushed commit `b0dfc72` to `nodefect/conventions:refactor-newtree-git-c-fullpath`.

**What changed:**

- **Removed** the helper functions `GitPointerFileInRepo` and `BareRepoDirInRepo`.
- **Inlined** their bodies at all three callsites so each location can independently decide whether to pass `checkExistence = true/false` (or accept the default). All current callsites keep `checkExistence = false` because:
  1. When validating an existing clone, `.git`/`.bare` may legitimately not exist (and the code checks `.Exists` manually to emit a custom error).
  2. When creating the `.git` pointer file after a fresh clone, the file does not exist yet before `WriteAllText` is called.

This removes the hidden assumption baked into the helpers and lets future callers choose the appropriate behavior at each callsite.

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/27621505609) | Model: ppq/autoclaw | Time: 1m 56s | Tokens: 289.0K | Cost: $0.10 | Pi SDK v0.78.1 | Action v2.19.2

knocte:
/pi I think I see duplicated code:

```
            let gitFile =
                cloneDir.CombineFile(
                    gitPointerFileName,
                    checkExistence = false
                )

            let bareDir =
                cloneDir.CombineDir(
                    bareRepoDirName,
                    checkExistence = false
                )

            if not gitFile.Exists || not bareDir.Exists then
                let exitCode, errMsg = ErrDirectoryIsNotClone firstArg
                Console.Error.WriteLine errMsg

                Environment.Exit exitCode
```

If this is indeed dupe, let's create a func and call it twice. Also scan for more other possible code fragments that might be duplicated, and do the same.

After you have worked on this in nodefect's fork (the branch of this PR), you can push directly to it because the TOKEN you have available corresponds to this nodefect github user account.

nodefect:
Done. I extracted two helpers and replaced all duplicated occurrences:

1. **`ValidateIsClone`** – encapsulates the repeated `.git`/`.bare` existence validation. Called from both the URL branch and the folder-name branch.

2. **`ExitWithError`** – encapsulates the repeated 4-line pattern of `let exitCode, errMsg = …`, `Console.Error.WriteLine errMsg`, `Environment.Exit exitCode`, plus the unreachable `failwith` that follows it in some branches. Replaced **15 occurrences** across the file (including inside `ResolvePrUrl`, the clone-failure handler, and the remote-name determination logic).

Pushed as commit `72676c0` to `nodefect/conventions:refactor-newtree-git-c-fullpath`.

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/27622260712) | Model: ppq/autoclaw | Time: 10m 12s | Tokens: 1.8M | Cost: $0.62 | Pi SDK v0.78.1 | Action v2.19.2

scripts/newTree.fsx
```diff
@@ -148,6 +148,11 @@ let ErrFailedToExtractHeadRepoUrl exMsg =
 let ErrFailedToExtractHeadBranch exMsg =
     (14, sprintf "Failed to extract head branch from PR data: %s" exMsg)
 
+let ExitWithError (exitCode, errMsg): 'a =
```
knocte: let's remove `'a` here

PR #320, opened 2026-06-17
refactor(scripts/newTree.fsx): exit codes

knocte:
* Unexpected things should just fail with an exception, no need for error code ceremony.
* Things that fail very likely because of internet connection issues should have this as a hint in error msgs.
* References to directory name should have full path, to be less ambiguous.

### DIFF
--- a/scripts/newTree.fsx
+++ b/scripts/newTree.fsx
@@ -113,22 +113,24 @@ let ErrDirectoryDoesNotExist path =
     (2, sprintf "Directory '%s' does not exist." path)
 
 let ErrDirectoryIsNotClone path =
-    (3, sprintf "Directory '%s' already exists and is not a clone." path)
+    (3, sprintf "Directory '%s' already exists but is not a clone." path)
 
 let errCannotDetermineRemoteName =
     (4,
      "Both 'upstream' and 'origin' remotes already exist and the new remote is not a fork. Cannot determine a name for the new remote.")
 
-let errCannotCheckGitHubFork =
+let errGitHubApiConnectionFailedCannotCheckIfGitHubRepoIsForkOrNot =
     (5,
-     "Could not check whether the repo is a GitHub fork. Cannot determine a name for the new remote.")
+     "Could not check whether the repo is a GitHub fork: check your internet connection?")
 
-let ErrFailedToAddRemote url =
-    (6, sprintf "Failed to add remote '%s'." url)
+let errPrAlreadyEncodesBranchName =
+    (6,
+     "A PR URL already encodes the branch name. Do not specify a second argument.")
 
-let errGitFetchAllFailed = (7, "Git fetch --all failed.")
+let errGitFetchAllFailed =
+    (7, "Git fetch --all failed: check your internet connection?")
 
-let errGitCloneFailed = (8, "Git clone failed.")
+let errGitCloneFailed = (8, "Git clone failed: check your internet connection?")
 
 let ErrNotGitHubCannotDetermineRemoteName dir =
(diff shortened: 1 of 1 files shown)


nodefect:
**Overall assessment: positive refactor, aligns well with stated goals. Minor suggestions for consistency.**

---

### ✅ What works well

**1. Removing unnecessary error-code ceremony for unexpected failures**
The removal of `ErrFailedToAddRemote`, `ErrFailedToExtractHeadRepoUrl`, and `ErrFailedToExtractHeadBranch` is a good call. If GitHub's API returns malformed JSON or `git remote add` fails, the underlying exception message is usually more actionable than a generic wrapped string. The `Environment.Exit` + `failwith "Unreachable"` pattern was indeed cerimonial for cases that should never happen in normal operation.

**2. Network-related errors now hint at connection issues**
The rewording of:
* `errCannotCheckGitHubFork` → `errGitHubApiConnectionFailedCannotCheckIfGitHubRepoIsForkOrNot` ("check your internet connection?")
* `errGitFetchAllFailed` ("check your internet connection?")
* `errGitCloneFailed` ("check your internet connection?")
* `errFailedToFetchPrData` ("check your internet connection?")

…is helpful UX. These are the exact failure modes where a flaky connection is the most likely culprit.

**3. `ErrDirectoryDoesNotExist` now uses the full path**
Changing `firstArg` to `cloneDir.FullPath` removes ambiguity when the user passes a relative path like `../foo` or just `foo`. Good improvement.

**4. New dedicated error code for PR+branch misuse**
Formalizing the ad-hoc `"A PR URL already encodes the branch name..."` message into `errPrAlreadyEncodesBranchName` (reusing exit code 6 after `ErrFailedToAddRemote` was removed) is cleaner than reusing `errUsage` and printing a hardcoded string.

---

### ⚠️ Suggestions / things to consider

**a) Full-path goal isn't applied consistently**  
The PR description says *"References to directory name should have full path, to be less ambiguous."* This was done for `ErrDirectoryDoesNotExist`, but two other similar error sites still use short names:

* `ValidateDirIsClone` still passes `cloneDir.PathlessName` to `ErrDirectoryIsNotClone`
* `ErrNotGitHubCannotDetermineRemoteName` is still called with `repoAndFolderName` instead of `cloneDir.FullPath`

Since `cloneDir` is in scope at both locations, consider updating these as well for consistency with the stated goal.

**b) Exit-code renumbering**  
`errInvalidPrUrl` shifted from `11` → `10`, and `errFailedToFetchPrData` from `12` → `11`. Not a problem if nothing externalDepends on these codes, but worth noting for anyone parsing exit codes in wrapper scripts.

**c) Deleted-fork PR edge case**  
If `head.repo` is `null` (e.g. PR opened from a fork that was later deleted), the old code caught the exception and printed `"Failed to extract head repo URL from PR data: ..."`. The new code will let FSharp.Data throw directly. This is acceptable given the PR's explicit goal of *"unexpected things should just fail with an exception"*, but the resulting stack trace may be slightly less user-friendly than the old message.

---

### Verdict

**Approve with minor suggestions.** The refactor is sound. If you want to address point (a) in a follow-up, both `ErrDirectoryIsNotClone` and `ErrNotGitHubCannotDetermineRemoteName` could likewise be passed `cloneDir.FullPath` to fully satisfy the "full path" goal.

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/27673877158) | Model: ppq/autoclaw | Time: 8m 28s | Tokens: 57.7K | Cost: $0.05 | Pi SDK v0.78.1 | Action v2.19.2

PR #321, opened 2026-06-17
ci: use Docker + ghcr instead of installing deps in workflows

knocte:
Following the same strategy as fp-sdk:
- A Dockerfile that installs all system dependencies.
- Docker image gets labeled via MD5sum from Dockerfile; if it doesn't exist in ghcr.io yet, it uploads it and deletes the old ones.
- Before running CI, it retrieves the docker image from ghcr.io.

This makes CI much faster because we avoid apt update/install on every run.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -19,32 +19,83 @@ env:
   CHECKOUT_ACTION_VERSION: "4"
 
 jobs:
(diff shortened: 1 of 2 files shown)


knocte:
Superseded by https://github.com/tarsgate/conventions/pull/319

PR #327, opened 2026-06-18
scripts/checkCommits1by1.fsx: fix script

webwarrior-ws:
Lately the result of Github API check suites endpoint [1] has been returning results with few first entries having latest_check_runs_count=0, status=completed, conclusion=failure, created_at=updated_at. These results caused false positives, so filtering by latest_check_runs_count was added.

[1] https://docs.github.com/en/enterprise-server@3.20/rest/checks/suites?apiVersion=2022-11-28#list-check-suites-for-a-git-reference

### DIFF
--- a/scripts/checkCommits1by1.fsx
+++ b/scripts/checkCommits1by1.fsx
@@ -1401,6 +1401,9 @@ prCommits
             checkSuitesParsedJson.CheckSuites
             // discard check suites for the commit that are not from PR branch
             |> Seq.filter(fun suite -> suite.PullRequests.Length > 0)
+            // check suites with latest_check_runs_count=0 have also created_at=updated_at and conclusion=failure;
+            // they cause false positives, so ignore them
+            |> Seq.filter(fun suite -> suite.LatestCheckRunsCount > 0)
             // take the first one (triggered by push); second one is triggered by pull_request
             |> Seq.tryHead
 


nodefect:
This fix is correct and appropriately targeted.

**What the change does**
The diff adds one filter step to the pipeline that processes GitHub check suites. Before taking the first suite from the API response, it now excludes entries with `LatestCheckRunsCount = 0`. These entries have been returning from the endpoint as bogus completed/failure results (with `created_at == updated_at`), which caused the script to report false positives.

**Why it makes sense**
- A check suite with zero check runs carries no actionable CI status information, regardless of whether it is a GitHub-side artifact or a transient state.
- The placement is correct: it comes after filtering for suites that belong to the PR (`suite.PullRequests.Length > 0`) and before `tryHead`, so it only affects the selection of the relevant suite to evaluate.
- The comment clearly explains the motivation for future readers.

**Nit (non-blocking)**
The comment says "runs with `latest_check_runs_count=0`…" — "check suites" would be slightly more precise than "runs" since the sequence contains `CheckSuites`. This is minor and can be addressed in a follow-up or ignored.

Approved; the change is safe to merge.

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/27755939951) | Model: ppq/autoclaw | Time: 1m 40s | Tokens: 17.2K | Cost: $0.01 | Pi SDK v0.78.1 | Action v2.19.2

knocte:
> The comment says "runs with latest_check_runs_count=0…" — "check suites" would be slightly more precise than "runs" since the sequence contains CheckSuites. This is minor and can be addressed in a follow-up or ignored.

@webwarrior-ws address this?

webwarrior-ws:
> > The comment says "runs with latest_check_runs_count=0…" — "check suites" would be slightly more precise than "runs" since the sequence contains CheckSuites. This is minor and can be addressed in a follow-up or ignored.
> 
> @webwarrior-ws address this?

Replaced "runs" with "check suites" in the comment.

PR #329, opened 2026-06-18
scripts/checkCommits1by1.fsx: wait for status

webwarrior-ws:
When checking CI statuses of commits, wait until CI workflow has finished. Otherwise there may be false negatives because checkCommits1by1.fsx runs while other workflows are still running.


webwarrior-ws:
Checked for cases when 1 of the commits has no CI status and when 1 of the commits has red status. Failure in both cases, as expected. I'll squash the commits now.

PR #331, opened 2026-06-29
scripts/checkCommits1by1.fsx: ignore commits not in PR

webwarrior-ws:
When checking for PR statues, don't include commits in the PR branch after current PR's head commit. Doing that could lead to false positives.

### DIFF
--- a/scripts/checkCommits1by1.fsx
+++ b/scripts/checkCommits1by1.fsx
@@ -1290,6 +1290,8 @@ repository or organization and click on Actions button, then select General. Fro
 
     result
 
+let prHeadCommitHash = parsedJsonObj.PullRequest.Head.Sha
+
 let prCommits =
     let url = parsedJsonObj.PullRequest.Links.Commits.Href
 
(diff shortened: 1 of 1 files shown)


scripts/checkCommits1by1.fsx
```diff
@@ -1290,15 +1290,22 @@ repository or organization and click on Actions button, then select General. Fro
...
 let prCommits =
     let url = parsedJsonObj.PullRequest.Links.Commits.Href
 
     let prCommitsJsonString = GitHubApiCall url
 
     let parsedPrCommitsJsonObj = PRCommitsType.Parse prCommitsJsonString
-
-    parsedPrCommitsJsonObj
-    |> Seq.map(fun commit -> (commit.Sha, commit.Commit.Message))
+    match parsedPrCommitsJsonObj |> Array.tryFindIndex (fun commit -> commit.Sha = prHeadCommitHash) with
```
knocte: @webwarrior-ws can use Seq. instead of Array.?

webwarrior-ws: Changed to Seq.

PR #334, opened 2026-07-01
WIP0

knocte:
TODO:
* change CI to use Dockerfile.base instead of Dockerfile
* change CI to generate additional image for Dockerfile.extras (with same approach: only upload if it has changed)


knocte:
@webwarrior-ws fixed CI in https://github.com/tarsgate/conventions/pull/337 , I guess the branch had to be inside the same repo, not in a fork

PR #336, opened 2026-07-06
scripts,Test: upgrade Fsdk to latest version

webwarrior-ws:
Because it has workaround for `preferreduilang` argument appearing in args.


webwarrior-ws:
Already included in https://github.com/tarsgate/conventions/pull/339 (merged)

PR #338, opened 2026-07-08
Copy github action to separate directory

webwarrior-ws:
So when referenced by external repositories, it's more clear what the action does, as reference will include path to the action.

Original action.yml at the root is not deleted to not disrupt current users. After this gets merged into master, original file can be deleted.

### DIFF
--- a/.github/actions/dynamic-uses/action.yml
+++ b/.github/actions/dynamic-uses/action.yml
@@ -1,0 +1,43 @@
+name: Dynamic Uses
+description: Dynamically resolve and use another GitHub action. Workaround for https://github.com/actions/runner/issues/895
+author: Jon Jensen
+inputs:
+  uses:
+    description: Action reference or path, e.g. `actions/setup-node@v3`
+    required: true
+  with:
+    description: 'JSON-ified `inputs` for the action, e.g. `{"node-version": "18"}`'
+    required: false
+    default: "{}"
+outputs:
+  outputs:
+    description: 'JSON-ified `outputs` from the action, e.g. `{"node-version": "v18.12.0", "cache-hit": true}`'
+    value: ${{ steps.run.outputs.outputs }}
+runs:
+  using: composite
+  steps:
+    - name: Setup
+      shell: bash
+      run: |
+        mkdir --parents ./.tmp-dynamic-uses &&
+        cat <<'DYNAMIC_USES_EOF' >./.tmp-dynamic-uses/action.yml
+        outputs:
+          outputs:
+            value: ${{ '$' }}{{ toJSON(steps.run.outputs) }}
+        runs:
+          using: composite
+          steps:
+          - run: rm --recursive --force ./.tmp-dynamic-uses
+            shell: bash
+          - name: Run
+            id: run
+            uses: ${{ inputs.uses }}
+            with: ${{ inputs.with || '{}' }}
+        DYNAMIC_USES_EOF
+    - name: Run
+      id: run
+      uses: ./.tmp-dynamic-uses
+    - name: Cleanup
+      if: always() && steps.run.outcome != 'success'
+      shell: bash
+      run: rm --recursive --force ./.tmp-dynamic-uses


knocte:
Mention action.yml in commit msg pls.

webwarrior-ws:
> Mention action.yml in commit msg pls.

It is already mentioned in the commit message.

knocte:
> So when referenced by external repositories

Are we actually referencing it from external repos? I see we stopped using it in pulumi-deploy because we stopped using the pulumi action.

webwarrior-ws:
> > So when referenced by external repositories
> 
> Are we actually referencing it from external repos? I see we stopped using it in pulumi-deploy because we stopped using the pulumi action.

Yes, we stopped using it from pulumi-deploy.

knocte:
So then this doesn't need to be 2 PRs, just one is enough, right?

webwarrior-ws:
> So then this doesn't need to be 2 PRs, just one is enough, right?

No, because this repository references the action as `nblockchain/conventions@master`.
Alternatively, it can be referenced as local action, in that case 1 PR should be enough.

.github/actions/dynamic-uses/action.yml
```diff
@@ -0,0 +1,43 @@
+name: Dynamic Uses
```
overtongeist: Verified: this is a byte-for-byte identical copy of the root `action.yml`. The location `.github/actions/dynamic-uses/` follows the standard GitHub layout for reusable actions, and the directory name mirrors the action's `name: Dynamic Uses`, so the external reference `owner/conventions/.github/actions/dynamic-uses@master` is self-describing — exactly the goal stated in the PR description. 👍

.github/actions/dynamic-uses/action.yml
```diff
@@ -0,0 +1,43 @@
...
+  outputs:
+    description: 'JSON-ified `outputs` from the action, e.g. `{"node-version": "v18.12.0", "cache-hit": true}`'
+    value: ${{ steps.run.outputs.outputs }}
+runs:
+  using: composite
+  steps:
+    - name: Setup
+      shell: bash
+      run: |
+        mkdir --parents ./.tmp-dynamic-uses &&
```
overtongeist: Non-blocking, pre-existing (verbatim copy of the root action, not introduced by this PR): these GNU long options (`--parents`, and `--recursive --force` further down) don't exist on macOS/BSD runners, so the action only works on Linux runners. Fine given current usage — just flagging it as a caveat for external consumers who might try it on a `macos-*` runner.

overtongeist (approved):
## Review Summary

**What this PR does:** adds `.github/actions/dynamic-uses/action.yml` as a copy of the root `action.yml`, so external repositories can reference it as `owner/conventions/.github/actions/dynamic-uses@master` — a self-describing path — while the root file stays in place so current users aren't disrupted.

**Verification performed:**

- ✅ **Exact copy**: the new file is byte-for-byte identical to the root `action.yml` (verified with `diff` — no drift, no accidental edits).
- ✅ **Scope**: single file added, nothing else touched — matches the description.
- ✅ **Location & naming**: `.github/actions/<name>/action.yml` is the standard GitHub layout for reusable actions; the directory name matches the action's `name: Dynamic Uses`, so the external reference path now says what the action does.
- ✅ **Repo conventions**: file ends with EOL and is LF-only (passes `eofConvention` / `mixedLineEndings` checks); no unpinned action versions introduced (the inner `uses:` is dynamic, same as the original).
- ✅ **CI**: all 7 checks pass on the head commit (`c2e8f38`).

**Non-blocking notes / follow-ups:**

1. **Interim duplication**: until the root `action.yml` is deleted, any change to the action has to be mirrored in both copies — worth keeping that window short so they don't drift.
2. **Follow-up PR checklist** (for when the root file is eventually removed): the 5 references in `.github/workflows/CI.yml` (`uses: nblockchain/conventions@master`) point at the root action and must first be updated to the new path (`nblockchain/conventions/.github/actions/dynamic-uses@master`). Note they can't become local `./` references, because these steps run *before* `actions/checkout` — checkout itself is invoked via this action, which is the whole point of the runner#895 workaround.
3. **Pre-existing caveat** (copied verbatim, not introduced by this PR): the `Setup`/`Cleanup` steps use GNU long options (`mkdir --parents`, `rm --recursive --force`), which fail on macOS/BSD runners. Fine for Linux-only usage; just something external consumers should be aware of.

Overall: straightforward, verified, CI green — good to merge. 👍

overtongeist:
The repo is `tarsgate/conventions` (upstream self-references as `nblockchain/conventions`). I've pulled the PR branch and verified the change against the root `action.yml`, the repo's file conventions, and CI status. Posting the review:**Verdict: ✅ Approve** — [review posted](https://github.com/tarsgate/conventions/pull/338#pullrequestreview-5140470011) with 2 inline comments.

The PR does exactly what it claims. Key verifications:

- **Exact copy** — `.github/actions/dynamic-uses/action.yml` is byte-for-byte identical to the root `action.yml` (`diff` clean, no accidental edits).
- **Scope** — single file added, nothing else changed.
- **Conventions** — new file ends with EOL and is LF-only (passes `eofConvention` / `mixedLineEndings`); no unpinned action versions introduced.
- **Location** — `.github/actions/<name>/action.yml` is the standard layout for reusable actions, and `dynamic-uses` mirrors the action's name, so the external reference `owner/conventions/.github/actions/dynamic-uses@master` is now self-describing.
- **CI** — all 7 checks pass on head commit `c2e8f38`.

Non-blocking follow-ups noted in the review:

1. Until the root `action.yml` is deleted, changes must be mirrored in both copies — keep the window short.
2. When removing the root file later, the 5 `uses: nblockchain/conventions@master` references in `.github/workflows/CI.yml` must first be pointed at the new path (they can't be local `./` references since they run before checkout — which itself goes through this action).
3. Pre-existing caveat copied along verbatim: `mkdir --parents` / `rm --recursive --force` are GNU-only flags, so the action won't work on macOS runners — worth knowing for external consumers.

---

[View action run](https://github.com/tarsgate/conventions/actions/runs/34214841623) | Model: ppq/autoclaw | Time: 2m 41s | Tokens: 159.7K | Cost: $0.03 | Pi SDK v0.84.3 | Action v2.27.1

knocte:
> because this repository references the action as nblockchain/conventions@master.

How will that change after we remove action.yml ? Sorry I'm revisiting this and wrapping my head around the need for 2 PRs again.

knocte:
> How will that change after we remove action.yml ? Sorry I'm revisiting this and wrapping my head around the need for 2 PRs again.

Nevermind, merged. @webwarrior-ws please create new PR, commit msg title should start with "[2/2]"

PR #339, opened 2026-07-09
Upgrade FSharpLint

webwarrior-ws:
New version of FSharpLint has several bugs fixed that affected self-check in pulumi-deploy repository.

### DIFF
--- a/fsharplint.json
+++ b/fsharplint.json
@@ -333,6 +333,35 @@
         }
     },
     "ensureTailCallDiagnosticsInRecursiveFunctions": { "enabled": true },
+    "favourAsKeyword": { "enabled": true },
+    "interpolatedStringWithNoSubstitution": { "enabled": true },
+    "indexerAccessorStyleConsistency": {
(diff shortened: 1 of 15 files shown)


knocte:
Please squash 1st and 2nd commit, no need to have them separated, no point at all.
Also, Fsdk commit should be the last, to have one less red CI status.

knocte:
Why is CI red?

knocte:
CI still can be fixed

knocte:
CI is **still** red

webwarrior-ws:
> CI is **still** red

Green now

knocte:
Please create FSharpLint PR that:
* Adds unit test that shows the false positive for the hint that you eliminated from here.
* Second commit: revert the commit that adds the hint.
* Third commit: marks test as Ignored because fixing all hashcode edge cases is not worth it.

knocte:
> Please create FSharpLint PR that:

PR871 created.

